PR #36, opened 2022-01-07
Feature/layers re-architecture

jakobuid:
The goal of this PR is to provide the project with the structured layers/modules. Each module is divided into the defined packages structure. Also, the new naming convention of models in each layer is defined.


sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
```diff
@@ -1,18 +1,18 @@
...
+import com.walletconnect.walletconnectv2.client.Init
+import com.walletconnect.walletconnectv2.client.WalletConnectClient
 
 class SampleApplication : Application() {
 
     override fun onCreate() {
         super.onCreate()
 
-        val initParams = ClientTypes.InitialParams(
+        val init = Init(
```
TalhaAli00, line 13: So the `Init` object doesn't require adding the `WalletConnectClient` in front of it, should we look into another solution so developers know that the `Init` object is for WC?

sample/src/main/java/com/walletconnect/sample/wallet/SessionActionListener.kt
```diff
@@ -1,11 +1,11 @@
...
-    fun onDisconnect(session: WalletConnectClientData.SettledSession)
-    fun onUpdate(session: WalletConnectClientData.SettledSession)
-    fun onUpgrade(session: WalletConnectClientData.SettledSession)
-    fun onPing(session: WalletConnectClientData.SettledSession)
-    fun onSessionsDetails(session: WalletConnectClientData.SettledSession)
+    fun onDisconnect(session: SettledSession)
+    fun onUpdate(session: SettledSession)
+    fun onUpgrade(session: SettledSession)
+    fun onPing(session: SettledSession)
+    fun onSessionsDetails(session: SettledSession)
```
TalhaAli00, line 10: Same comment as above about having the `WalletConnectClient` prefix i.e. WalletConnectClient.SettledSession

sample/src/main/java/com/walletconnect/sample/wallet/WalletFragment.kt
```diff
@@ -85,23 +85,23 @@ class WalletFragment : Fragment(R.layout.wallet_fragment), SessionActionListener
...
+    override fun onUpgrade(session: SettledSession) {
         viewModel.sessionUpgrade(session)
     }
 
-    override fun onPing(session: WalletConnectClientData.SettledSession) {
+    override fun onPing(session: SettledSession) {
         viewModel.sessionPing(session)
     }
 
-    override fun onSessionsDetails(session: WalletConnectClientData.SettledSession) {
```
TalhaAli00, line 104: same comment as above

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -0,0 +1,379 @@
...
+        //TODO delete from DB
+        return EngineDO.Default
+    }
+
+    private fun onPing(topic: TopicVO, requestId: Long): EngineDO.Default {
+        val jsonRpcResult = EngineDO.JsonRpcResponse.JsonRpcResult(id = requestId, result = "true")
+        relayer.respond(topic, jsonRpcResult.toJsonRpcResponseVO(),
+            { Logger.log("Ping send successfully") },
+            { error -> Logger.error("Ping Error: $error") })
+        return EngineDO.Default
```
TalhaAli00, line 302: I thought we agreed that we were going to move these methods to the relayer? was there a problem with that idea?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/network/NetworkRepositoryInterface.kt
```diff
@@ -0,0 +1,3 @@
+package com.walletconnect.walletconnectv2.network
+
+interface NetworkRepositoryInterface
```
TalhaAli00, line 3: I think this does not need the word `Interface` in the name

sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
jakobuid: Yea we can go back to the sealed class solution let's say called `WalletConnectParams`. Let me know what you think.

sample/src/main/java/com/walletconnect/sample/wallet/SessionActionListener.kt
jakobuid: Same comment

sample/src/main/java/com/walletconnect/sample/wallet/WalletFragment.kt
jakobuid: Same comment

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
jakobuid: Yes because because what those methods do is actually the Engine responsibility.
Relay:
- Handles the peers communications
- Publish the requests
- Manage subscriptions
- Decrypts the incoming messages and deserialise them
- Encrypts the outgoing messages and serialise them

Engine:
- Handles the sequence storage
- Constructs data that has to be send on particular method
- Handles the behavioural logic of each method
- Validates the incoming and outgoing messages
- Validates the permissions

Those methods are not yet fully implemented. E.g. the validation of messages and permissions will be placed there, that's why imo it should stay in the Engine.

sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
jakobuid: after the discussion we agreed to create one WalletConnect sealed class with Models, Listeners and Params

TalhaAli00:
```suggestion
        object WalletConnect {
           sealed interface Listeners {
              fun onError(error: Throwable)
           }

           sealed class Model

           sealed class Params {
               companion object {
                   internal const val WALLET_CONNECT_URL = "relay.walletconnect.com"
               }
           }
        }
```

TalhaAli00 (approved):

PR #48, opened 2022-01-18
added Kodein as DI library. Also added modifiers to classes and funct…

TalhaAli00:
…ions to try to reduce visibility in Kotlin and Java files


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/di/DITags.kt
```diff
@@ -0,0 +1,9 @@
+@file:JvmSynthetic
+
+package com.walletconnect.walletconnectv2.di
+
+internal enum class DITags {
+    @JvmSynthetic KEYSTORE,
+    @JvmSynthetic SERVER_URL,
+    @JvmSynthetic RPCSTORE
```
jakobuid, line 8:
```suggestion
    @JvmSynthetic RPC_STORE
```

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -31,29 +31,19 @@ import kotlinx.coroutines.flow.*
...
-    private val crypto: CryptoRepository = BouncyCastleCryptoRepository()
-    //endregion
-
-    private var metaData: EngineDO.AppMetaData? = null
-    private var controllerType = ControllerType.CONTROLLER
-
+internal class EngineInteractor(
+    private val relayer: WalletConnectRelayer,
+    private val crypto: CryptoRepository,
+    private val sequenceRepository: SequenceStorageRepository,
```
jakobuid, line 37:
```suggestion
    private val sequenceStorageRepository: SequenceStorageRepository,
```

jakobuid (approved):
💪

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/common/model/vo/clientsync/pairing/PairingParamsVO.kt
```diff
@@ -49,28 +49,28 @@ internal sealed class PairingParamsVO : ClientParams {
...
 
     @JsonClass(generateAdapter = true)
-    class DeleteParams(
+    internal class DeleteParams(
         @Json(name = "reason")
         val reason: ReasonVO
     ) : PairingParamsVO()
 
-    class PingParams : PairingParamsVO()
+    internal object PingParams : PairingParamsVO()
```
jakobuid, line 63: Here we should use `class` instead of `object` because there is a crash when we send a ping to peer.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/common/model/vo/clientsync/session/SessionParamsVO.kt
```diff
@@ -53,35 +53,35 @@ internal sealed class SessionParamsVO: ClientParams {
...
     ) : SessionParamsVO()
 
-    data class SessionPermissionsParams(
+    internal data class SessionPermissionsParams(
         @Json(name = "permissions")
         val permissions: SessionPermissionsVO
     ) : SessionParamsVO()
 
-    class PingParams : SessionParamsVO()
+    internal object PingParams : SessionParamsVO()
```
jakobuid, line 82: Here we should use `class` instead of `object` because there is a crash when we send a ping to peer.

jakobuid (changes requested):

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/common/scope/WalletConnectScope.kt
```diff
@@ -1,41 +1,11 @@
-@file:JvmName("WalletConnectScope")
+@file:JvmName(" WalletConnectScope")
```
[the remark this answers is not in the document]
TalhaAli00: so adding special characters is a way to hide the file from Java so this was intentional

jakobuid (approved):

PR #50, opened 2022-01-19
Feature/add require init logic

TalhaAli00:


jakobuid (approved):
It can be merged on my end.
However, please remember to put **only** changes into to the PR that applies to given task.  From the reviewer perspective it's super helpful.

TalhaAli00 (commented):
I agree with your comment. The only reason the changes were showing more than what was changed was because it was branched from the DI branch. If you want to refresh now, you'll see the change

PR #60, opened 2022-01-25
Feature/update all public proposer methods

jakobuid:
This piece of code is responsible for adding the missing proposer implementation of all public methods.


sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
```diff
@@ -28,7 +28,7 @@ class SampleApplication : Application() {
         val init2 = WalletConnect.Params.Init(
             application = this,
             relayServerUrl = "wss://$WALLET_CONNECT_URL/?projectId=$PROJECT_ID",
-            isController = true,
+            isController = false,
```
TalhaAli00, line 31: this is the wallet's init params. Shouldn't this be true for wallets?

sample/src/main/java/com/walletconnect/sample/dapp/DappFragment.kt
```diff
@@ -3,21 +3,51 @@ package com.walletconnect.sample.dapp
...
         binding = DappFragmentBinding.bind(view)
         setupToolbar()
+
+        //TODO: Improve the dApp sample app. Code below is only for test purposes!
+
+        val connectParams = WalletConnect.Params.Connect(
+            WalletConnect.Model.SessionPermissions(
+                WalletConnect.Model.Blockchain(listOf("eip155:69")),
+                WalletConnect.Model.Jsonrpc(listOf("personal_sign"))
+            ), null
```
TalhaAli00, line 32: use named parameters

sample/src/main/java/com/walletconnect/sample/wallet/WalletViewModel.kt
```diff
@@ -23,7 +24,7 @@ class WalletViewModel : ViewModel(), WalletConnectClient.WalletDelegate {
     fun pair(uri: String) {
         val pair = WalletConnect.Params.Pair(uri.trim())
         WalletConnectClient.pair(pair, object : WalletConnect.Listeners.Pairing {
-            override fun onSuccess(settledPairing: WalletConnect.Model.SettledPairing) {
+            override fun onSuccess(settledPairing: WalletConnect.Model.SettledSequence) {
```
TalhaAli00, line 27: looking at this being used, why changed from `SettledPairing` to `SettledSequence` when it's being used for the pair function

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnect.kt
```diff
@@ -100,7 +103,12 @@ object WalletConnect {
 
         data class SessionState(val accounts: List<String>) : Model()
 
-        data class SettledPairing(val topic: String) : Model()
+        data class SettledPairing(
+            val topic: String,
+            val appMetaData: AppMetaData?
+        ) : Model()
+
+        data class SettledSequence(val topic: String) : Model()
```
TalhaAli00, line 111: why can't these be consolidated?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnectClient.kt
```diff
@@ -226,4 +255,20 @@ object WalletConnectClient {
...
+        }
+    }
+
+    private fun safeCallWithReturn(method: () -> String?): String? {
+        try {
+            return method()
+        } catch (e: WalletConnectExceptions) {
+            throw WalletConnect.WalletConnectError(e.message)
+        }
+    }
```
TalhaAli00, line 273: I don't think these are really needed. If there's an exception, they're just wrapping the exception's message with a WC error so we're exchanging one exception for another. Why can we not just expose `WalletConnectExceptions`? Or maybe make `WalletConnectExceptions` into `WalletConnect.WalletConnectError` and use that globally in the app

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/mapper/ClientMapper.kt
```diff
@@ -86,10 +86,7 @@ internal fun EngineDO.SessionApproved.toClientSessionApproved(): WalletConnect.M
 
 @JvmSynthetic
 internal fun WalletConnect.Model.SessionPermissions.toEngineSessionPermissions(): EngineDO.SessionPermissions =
-    EngineDO.SessionPermissions(
-        blockchain.chains.let { chains -> EngineDO.Blockchain(chains) },
-        jsonRpc.methods.let { methods -> EngineDO.JsonRpc(methods) }
-    )
+    EngineDO.SessionPermissions(EngineDO.Blockchain(blockchain.chains), EngineDO.JsonRpc(jsonRpc.methods))
```
TalhaAli00, line 89: nice

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/common/model/vo/clientsync/session/SessionParamsVO.kt
```diff
@@ -53,7 +53,10 @@ internal sealed class SessionParamsVO: ClientParams {
     ) : SessionParamsVO()
 
     @JsonClass(generateAdapter = true)
-    internal class RejectParams(val reason: String) : SessionParamsVO()
+    internal class RejectParams(
+        @Json(name = "reason")
```
TalhaAli00, line 57: Why does a VO object need the moshi? Moshi should be related to the DTO models

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/model/EngineDO.kt
```diff
@@ -147,4 +151,6 @@ internal sealed class EngineDO {
             val message: String,
         )
     }
+
+    data class Request(val topic: String, val method: String, val params: String, val chainId: String?) : EngineDO()
```
TalhaAli00, line 155: add internal

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -59,34 +61,44 @@ internal class EngineInteractor(
     }
 
     internal fun proposeSequence(permissions: EngineDO.SessionPermissions, pairingTopic: String?): String? {
+        checkPeer(controllerType == ControllerType.NON_CONTROLLER) {
```
TalhaAli00, line 64: instead of passing the checking logic, you can just pass the `ControllerType` that is expected an do the boolean check inside `checkPeer`

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -59,34 +61,44 @@ internal class EngineInteractor(
...
+            "The connect() was called by the unauthorized peer. Call it with: ${ControllerType.CONTROLLER}"
```
TalhaAli00, line 65: Instead of telling users to use the `ControllerType`, we need to explain that they need to re-initialize with the `isController` flag switched. Maybe we should use `ControllerType` in the init params instead of a `isController` flag

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -59,34 +61,44 @@ internal class EngineInteractor(
...
             val selfPublicKey: PublicKey = crypto.generateKeyPair()
             val isController = controllerType == ControllerType.CONTROLLER
-            val proposalParams = SessionProposerVO(selfPublicKey.keyAsHex, isController, metaData.toClientSyncMetaData())
+            val proposalParams = SessionProposerVO(selfPublicKey.keyAsHex, isController, metaData.toMetaDataVO())
                 .toProposalParams(pendingSessionTopic, settledPairing.topic, permissions)
 
             val proposedSession: SessionVO = proposalParams.toProposedSessionVO(pendingSessionTopic, selfPublicKey, controllerType)
-            sequenceStorageRepository.insertSessionProposal(proposedSession, metaData.toClientSyncMetaData(), controllerType)
+
+            Logger.error("Inserting the session proposal: $proposedSession")
```
TalhaAli00, line 92: why are we logging an error here?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -430,10 +555,17 @@ internal class EngineInteractor(
...
 
-    internal fun getListOfSettledSessions(): List<EngineDO.SettledSession> =
-        sequenceStorageRepository.getListOfSessionVOs()
+    internal fun getListOfSettledSessions(): List<EngineDO.SettledSession> {
+        return sequenceStorageRepository.getListOfSessionVOs()
             .filter { session -> session.status == SequenceStatus.ACKNOWLEDGED && session.expiry.isSequenceValid() }
             .map { session -> session.toEngineDOSettledSessionVO() }
+    }
+
+    fun getListOfSettledPairings(): List<EngineDO.SettledPairing> {
```
TalhaAli00, line 564: add internal

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -388,37 +492,58 @@ internal class EngineInteractor(
...
+            EngineDO.Default
+        }
         return params.toEngineDoSessionNotification(topic)
     }
 
-    private fun onPairingDelete(params: PairingParamsVO.DeleteParams, topic: TopicVO): EngineDO.Default {
+    private fun onPairingDelete(params: PairingParamsVO.DeleteParams, topic: TopicVO): EngineDO.DeletedPairing {
+        if (!sequenceStorageRepository.hasPairingTopic(topic)) {
+            Logger.error("onPairingDelete: No pairing for topic: $topic")
+            EngineDO.Default
```
TalhaAli00, line 534: Instead of `EngineDO.Default`, can we use something more descriptive like `EngineDO.NoPairing` or something?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -388,37 +492,58 @@ internal class EngineInteractor(
...
     }
 
     private fun onSessionNotification(params: SessionParamsVO.NotificationParams, topic: TopicVO): EngineDO.SessionNotification {
-        //TODO: validate notification
+        //TODO: Add Notification validation
+
+        if (!sequenceStorageRepository.hasSessionTopic(topic)) {
+            Logger.error("onSessionNotification: No session for topic: $topic")
+            EngineDO.Default
+        }
```
TalhaAli00, line 527: same as below

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -388,37 +492,58 @@ internal class EngineInteractor(
...
         return params.toEngineDOSessionRequest(topic, requestId)
     }
 
     private fun onSessionDelete(params: SessionParamsVO.DeleteParams, topic: TopicVO): EngineDO.DeletedSession {
+        if (!sequenceStorageRepository.hasSessionTopic(topic)) {
+            Logger.error("onSessionDelete: No session for topic: $topic")
+            EngineDO.Default
+        }
```
TalhaAli00, line 513: Return something more descriptive like `EngineDO.FailedTopic`

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -298,17 +368,28 @@ internal class EngineInteractor(
...
                 onFailure = { error -> onFailure(error) }
             )
         }
     }
 
+    private fun checkPeer(isAuthorizedPeer: Boolean, message: () -> String) {
+        if (!isAuthorizedPeer) {
+            throw WalletConnectExceptions.UnauthorizedPeerException(message())
+        }
+    }
```
TalhaAli00, line 391: instead of taking a boolean value as a parameter, take a `ControllerType` and then just check inside the function

sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
jakobuid: Yea, I changed to false to test the proposer.

sample/src/main/java/com/walletconnect/sample/wallet/WalletViewModel.kt
jakobuid: The reason is that the metadata in SettledPairing is always null at this point, when the user receives the success when pairing. I agree that the original naming is better. I changed it back to the SettletPairing, with default value for metadata as null.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnectClient.kt
jakobuid: If we think of the separations of concerns, this approach is needed. We agreed that we cannon expose to the user models which are not in the client package. 

The  internal `WalletConnectExceptions` class is in the `common` package, so it cannot be exposed to the user.
Also, we cannot make `WalletConnectExceptions` into the `WalletConnect.WalletConnectError` and use globally in the app because it's in the client package.

When you think of our layers as modules, you cannot use a class in the client module as global, because the rest of the modules do not have a knowledge about this class.

We should map models that comes from engine/common package to client facing models.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/common/model/vo/clientsync/session/SessionParamsVO.kt
jakobuid: Most of the models under the `common/model/vo/clientsyns` have the moshi annotations. Those are used in the Engine to construct the Peer message sent by Relayer.
If the class field name matches the json name, we can remove those annotation.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
jakobuid: Yea I had this dylema. I couldn't decide which approach I like better. I don't really mind, I can change the boolean check to be done internally in the method :)

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
jakobuid: No, I don't want to change the public api. I will change the message to: ` "The ${method_name} was called by the unauthorized peer. Initialize SDK with isController = true flag`

Good call 👌

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
jakobuid: Leftovers from debugging

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
jakobuid: Comment above

sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
TalhaAli00: I think we should change the name of our init variables from `init` and `init2` to `initDapp` and `initWallet` so we don't have to change anything during tests

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnectClient.kt
TalhaAli00: I agree that with our current implementation this approach is needed. But an easy solution is to move the exceptions in a "core" directory which can become a core module. Everything in the core module would become global across the SDK and we can make it public if we want. That way there's only one type that we have to worry about.

I'm pushing for a single exception because in terms of modularization, it's going to become messy once we start having different SDK for our different layers. Imagine when we have a client SDK, Relayer SDK, and Auth SDK for example. With our current approach, each SDK will have it's own public facing exception and it's own mapping logic. We can simply this scenario with having global exceptions in a core module that are shared everywhere. Exceptions are different from data models, they are to express critical failures instead of SDK state

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -118,7 +125,14 @@ internal class EngineInteractor(
     }
 
     internal fun pair(uri: String, onSuccess: (String) -> Unit, onFailure: (Throwable) -> Unit) {
+        checkPeer(ControllerType.CONTROLLER) {
+            "The connect() was called by the unauthorized peer. Initialize SDK with isController = true."
```
TalhaAli00, line 129: maybe we should move these hardcoded strings to a file with global variables.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnectClient.kt
TalhaAli00, line 273: You can consolidate these into a single function with a generic return type

```suggestion
    private fun safeCall(method: () -> T): T {
        try {
            return method()
        } catch (e: WalletConnectExceptions) {
            throw WalletConnect.WalletConnectError(e.message)
        }
    }
```

sample/src/main/java/com/walletconnect/sample/SampleApplication.kt
jakobuid: Yes, good call.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnectClient.kt
jakobuid: I totally agree. Good call.
I think the best what we can do is to change the name of our `common` package to `core`, because imo it's better to have one global module than two. In the core module, all the components are with the `internal` keyword, except those which need to be publicly exposed.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
jakobuid: Nice improvement, I'll add such file to core package.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnectClient.kt
jakobuid: Since we agreed that exception goes to the core module and are used across all layers, it's not needed.

TalhaAli00 (approved):

PR #61, opened 2022-01-26
Feature/handle proposer internal calls

jakobuid:
The PR covers all missing Proposer's internal calls after the session settlement:

- onPairingUpdate
- onSessionUpdate
- onSessionUpgrade


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -572,10 +630,19 @@ internal class EngineInteractor(
...
+        }
+    }
+
+    private fun checkTopic(topic: TopicVO, onTopicNotFound: () -> Unit = {}): Boolean {
+        var isValid = true
+        if (!sequenceStorageRepository.hasSessionTopic(topic) || !sequenceStorageRepository.hasPairingTopic(topic)) {
+            onTopicNotFound()
+            isValid = false
         }
+        return isValid
```
TalhaAli00, line 645:
```suggestion
        val isValid = !sequenceStorageRepository.hasSessionTopic(topic) || !sequenceStorageRepository.hasPairingTopic(topic)
        
        if (isValid) {
            onTopicNotFound()
        }
        
        return isValid
```

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -572,10 +630,19 @@ internal class EngineInteractor(
...
 
-    private fun checkPeer(currentPeer: ControllerType, message: String) {
+    private fun checkPeer(currentPeer: ControllerType, onUnauthorizedPeer: () -> Unit) {
         if (controllerType != currentPeer) {
-            throw WalletConnectException.UnauthorizedPeerException(message)
+            onUnauthorizedPeer()
+        }
+    }
+
+    private fun checkTopic(topic: TopicVO, onTopicNotFound: () -> Unit = {}): Boolean {
```
TalhaAli00, line 639: instead of onTopicFound, have a second parameter of type String for the message. Inside this function, you'll call `Logger` and pass the string as the message. 
The third parameter can be a lamda function that returns T. Then everywhere you call `checkPeer`, instead of using `also,` just put the return logic in the lamda

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -572,10 +630,19 @@ internal class EngineInteractor(
...
+        var isValid = true
+        if (!sequenceStorageRepository.hasSessionTopic(topic) || !sequenceStorageRepository.hasPairingTopic(topic)) {
+            onTopicNotFound()
+            isValid = false
         }
+        return isValid
     }
```
TalhaAli00, line 646:
```suggestion
    private fun <T> checkTopic(topic: TopicVO, topicNotFoundMsg: String, invalidTopic: () -> T): T {
        val isValid = !sequenceStorageRepository.hasSessionTopic(topic) || !sequenceStorageRepository.hasPairingTopic(topic)
        
        if (!isValid) {
            Logger.error(errorMsg)
            invalidTopic()
        }
        
        return isValid
    }
```

jakobuid: I like your improvement but I slightly changed it because in some cases we don'r return but throw an error, so AD was showing warning.

TalhaAli00 (approved):

PR #77, opened 2022-02-11
Check expiry when fetching sequence from DB

jakobuid:
Check expiry when a sequence is fetched from DB. When expired delete all data coupled with sequence.

### DIFF
--- a/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/storage/sequence/SequenceStorageRepository.kt
+++ b/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/storage/sequence/SequenceStorageRepository.kt
@@ -12,6 +12,7 @@ import com.walletconnect.walletconnectv2.storage.data.dao.MetaDataDaoQueries
 import com.walletconnect.walletconnectv2.storage.data.dao.PairingDaoQueries
 import com.walletconnect.walletconnectv2.storage.data.dao.SessionDaoQueries
 import com.walletconnect.walletconnectv2.util.Empty
+import com.walletconnect.walletconnectv2.util.isSequenceValid
 
 //TODO: Split into SessionStorageRepository and PairingStorageRepository
 internal class SequenceStorageRepository(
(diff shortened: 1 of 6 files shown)


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/storage/sequence/SequenceStorageRepository.kt
```diff
@@ -20,6 +21,9 @@ internal class SequenceStorageRepository(
     private val metaDataDaoQueries: MetaDataDaoQueries
 ) {
 
+    @JvmSynthetic
+    var onSequenceExpired: (topic: TopicVO) -> Unit = {}
```
TalhaAli00, line 25: instead of making this a public variable, you can just pass a lamda into `isSessionValid` and `isPairingValid` which can then be passed into `verifyExpiry`. Then you just define a private function in the EngineInteractor and use the :: way to pass the function

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/storage/sequence/SequenceStorageRepository.kt
```diff
@@ -246,6 +268,14 @@ internal class SequenceStorageRepository(
         sessionDaoQueries.deleteSession(topic.value)
     }
 
+    private fun verifyExpiry(expiry: ExpiryVO, topic: TopicVO, deleteSequence: () -> Unit): Boolean {
```
TalhaAli00, line 271: personally, I think instead of wrapping the expiry in `ExpiryVO` inside the `isValid` functions, this function should just accept a Long and then do the wrapping here. It's a small change, but it decouples the ExpiryVO and the `isValid` functions as well as consolidates the wrapping into one location

TalhaAli00 (approved):

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/storage/sequence/SequenceStorageRepository.kt
jakobuid: Good call, I agree

PR #79, opened 2022-02-16
Feature/add timeouts

jakobuid:
Added timeouts:
- 30s for ping response
- 5mins for session request response

Refactor of handling the peer response.


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/core/exceptions/peer/PeerError.kt
```diff
@@ -1,28 +1,49 @@
...
+    data class UnauthorizedJsonRpcMethod(val method: String) : Error()
+    data class UnauthorizedNotificationType(val type: String) : Error()
+    data class UnauthorizedUpdateRequest(val sequence: String) : Error()
+    data class UnauthorizedUpgradeRequest(val sequence: String) : Error()
+    data class UnauthorizedMatchingController(val isController: Boolean) : Error()
+}
+
+enum class Sequence {
+    SESSION, PAIRING
 }
```
TalhaAli00, line 49: can we move this to core?

TalhaAli00 (approved):

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/core/exceptions/peer/PeerError.kt
jakobuid: Yup sure

PR #101, opened 2022-03-09
Feature/add dapp mobile linking

TalhaAli00:


dapp/src/main/kotlin/com/walletconnect/dapp/ui/host/DappSampleActivity.kt
```diff
@@ -0,0 +1,39 @@
...
+    override fun onDestroy() {
+        super.onDestroy()
+
+        DappDelegate.selectedSessionTopic?.let {
+            val disconnectParams = WalletConnect.Params.Disconnect(sessionTopic = it, reason = "shutdown", reasonCode = 400)
+            WalletConnectClient.disconnect(disconnectParams)
+        }
+
+        // re-init scope
+//        WalletConnectClient.shutdown()
```
jakobuid, line 36: If commented I suggest deleting this

dapp/src/main/kotlin/com/walletconnect/dapp/ui/selected_account/SelectedAccountFragment.kt
```diff
@@ -0,0 +1,106 @@
...
+                dialog.dismiss()
+            }
+            .show()
+    }
+
+    private fun sendRequestDeepLink(sessionRequestDeeplinkUri: Uri) {
+        try {
+            requireActivity().startActivity(Intent(Intent.ACTION_VIEW, sessionRequestDeeplinkUri))
+        } catch (exception: ActivityNotFoundException) {
+            // There is no app to handle deep link
```
jakobuid, line 96: Maybe we can log message here?

jakobuid (changes requested):
Code looks good.
Two comments:
1. Concern is about the rest json rpc methods, other than personal_sign
2. When user dapp sends session proposal over existing pairing and wallet reject we might hide progress dialog when session reject event is received.

dapp/src/main/kotlin/com/walletconnect/dapp/domain/Chains.kt
```diff
@@ -0,0 +1,153 @@
...
+import com.walletconnect.dapp.R
+
+private val ETH_CHAIN = "eip155"
+// Commented out other methods due to wallet not understanding request
+// TODO: Fix other RPC methods
+private val defaultEthMethods: List<String> = listOf(
+//    "eth_sendTransaction",
+    "personal_sign"
+//    "eth_signTypedData"
+)
```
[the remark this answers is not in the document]
TalhaAli00: I wasn't able to figure out why these methods aren't working so I left them out of this PR. I'm sure it's because the sample from the docs aren't working samples similar to personal_sign.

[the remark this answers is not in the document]
TalhaAli00: We can add more methods later once the other work is complete. But atm, we have a working dapp example

dapp/src/main/kotlin/com/walletconnect/dapp/ui/host/DappSampleActivity.kt
TalhaAli00: Sure, I'll remove this. This actually needs to be a TODO in the SDK

jakobuid (approved):

PR #140, opened 2022-04-27
Updated Dapp to with sending deeplinks. Wallet now opens up on deepli…

TalhaAli00:
…nk requests

### DIFF
--- a/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
+++ b/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
@@ -822,7 +822,6 @@ internal class EngineInteractor(
             }
 
         listOfValidSessions
-            .filter { session -> session.isAcknowledged }
             .onEach { session -> relayer.subscribe(session.topic) }
     }
 
--- a/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
+++ b/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
@@ -43,7 +43,7 @@ internal class EngineInteractor(
     private val sessionProposalRequest: MutableMap<String, WCRequestVO> = mutableMapOf()
 
     init {
-        resubscribeToSettledSequences()
+        resubscribeToSequences()
         setupSequenceExpiration()
         collectJsonRpcRequests()
         collectJsonRpcResponses()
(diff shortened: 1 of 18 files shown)


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -822,7 +822,6 @@ internal class EngineInteractor(
             }
 
         listOfValidSessions
-            .filter { session -> session.isAcknowledged }
```
jakobuid, line 825: I suggest removing the settle keyword from `resubscribeToSettledSequences`, `resubscribeToSettledPairings` and `resubscribeToSettledSession` because we do not have settle sate anymore.

jakobuid (changes requested):
Two things in terms of the user flow:

- When user sends the session request, the dialog on the Wallet side is shown twice. Looks like shared emits current and previous event for session request dialog.

- It would be more user-friendly to redirect user to Dapp whenever he approves or rejects session proposal and session requests. Now after approving a session user is in the Wallet without active session, because only after second switch, Dapp sends session settle acknowledgement and session is established in both peers. So I think we should redirect user to Dapp in these situations.

TalhaAli00 (commented):
@jakobuid in terms of your second comment account switching back to dapp. There really isn't a nice way of doing this. The only options atm are to either introduce another deep link bringing our total of URIs to 3 or finishing the activity. Also, do we force a user to switch on every approval?

I'm hesitant about introducing more URIs because there could be multiple dapps installed on a phone. If a use defaults the deeplink to a single dapp, then the URI from the wallet will always open that dapp. Using URIs for wallets is safer since a user would have less wallets than dapps for accounts. 
The problem with finishing the activity is the user then has to manually open the wallet again which feels like more steps than switching apps again. 

In my mind, the best user experience is to use a notification to allow the user to approve/reject. But this will take more time to develop

TalhaAli00:
@jakobuid I'm not seeing the behavior you mentioned in the first point

https://user-images.githubusercontent.com/3485185/165674340-a634037b-8732-43ad-a688-2c588d81589d.mp4

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
TalhaAli00: done

jakobuid:
@TalhaAli00 What you're showing me on your video is not related to session request at all ;) 
You're showing sending session proposal.


https://user-images.githubusercontent.com/23366251/165692838-5ea9fc66-9fb0-4bfd-89aa-f08947e4d629.mp4

jakobuid:
> 

Ya, I agree that introducing more deeplinks to force going back top dapp from wallet after accepting session proposal could create an overhead for the whole flow.

I'm trying to think from user perspective and with current flow user is not aware that actually this second switch (from wallet to dapp) is required to finish the session settlement. Thereby, user could end-up confused because once approving session proposal actually there is no active session in the wallet, unless user does this switch.

Maybe `moveTaskToBack(true)` is sth that could help.

jakobuid (approved):

Issue #139, opened 2022-04-26
Add RelayClient as a part of Init params


jakobuid:
Allow developers to add RelayClient as a parameter in the initialise method.

- [ ] create a wrapper for Events and for each object inside the Event class
- [ ] modify the sharedFlow to do a map operation and map from the scarlet to the relay events
- [ ] modify the Engine interactor to use relay events
- [ ] update the WalletConnect.Params.Init to allow passing nullable Relay and default to null
- [ ] update WalletConnectClient.Init to accept the defined relay
- [ ] update Relay koin module to either use defined relay or the default one

PR #142, opened 2022-04-29
Feature/Add Relay to Init Params

Elyniss:
Issue: https://github.com/WalletConnect/WalletConnectKotlinV2/issues/139

In order to add Relay to Init params, I had to expose as public a lot of network data classes and some VO data classes.  To see the issue revert to commit before exposing: [WIP: Before fixing encapsulation issues](https://github.com/WalletConnect/WalletConnectKotlinV2/commit/76d92d0c4061a7bae67293bd4f9f7c68397dce0b)

Let me know if this is acceptable or if we can achieve the goal without exposing too many classes.


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/network/model/RelayMessage.kt
```diff
@@ -0,0 +1,16 @@
+package com.walletconnect.walletconnectv2.network.model
+
+import com.tinder.scarlet.Message
+
+
+sealed class RelayMessage {
+    data class Text(val value: String) : RelayMessage()
+    class Bytes(val value: ByteArray) : RelayMessage() {
+        operator fun component1(): ByteArray = value
```
TalhaAli00, line 9: why the override?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/network/model/RelayShutdownReason.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import com.tinder.scarlet.ShutdownReason
+
+data class RelayShutdownReason(val code: Int, val reason: String) {
+    companion object {
+        private const val NORMAL_CLOSURE_STATUS_CODE = 1000
+        private const val NORMAL_CLOSURE_REASON = "Normal closure"
+
+        @JvmField
+        val GRACEFUL = RelayShutdownReason(NORMAL_CLOSURE_STATUS_CODE, NORMAL_CLOSURE_REASON)
```
TalhaAli00, line 11: doesn't look like it's used. Remove if unused

TalhaAli00 (commented):
instead of exposing our internal files as public (all of the VO and DO classes), we should create public facing classes inside `WalletConnect.Params` for each of these internal classes. Inside the RelayClient, we can have internal mapping extension functions that will map from `WalletConnect.Params` classes to our VO and DO classes

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/network/model/RelayMessage.kt
Elyniss: At first, I simply duplicated the Message class. However, upon reading I can't justify why it should be there. I've removed it.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnect.kt
```diff
@@ -151,6 +152,159 @@ object WalletConnect {
...
+                data class OnConnectionOpened<out WEB_SOCKET : Any>(val webSocket: WEB_SOCKET) :
+                    Event()
+
+                data class OnMessageReceived(val message: Message) : Event()
+                data class OnConnectionClosing(val shutdownReason: ShutdownReason) : Event()
+                data class OnConnectionClosed(val shutdownReason: ShutdownReason) : Event()
+                data class OnConnectionFailed(val throwable: Throwable) : Event()
+            }
+
+            sealed class Message : Model() {
```
TalhaAli00, line 301: Same here, Relay should be the parent

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnect.kt
```diff
@@ -151,6 +152,159 @@ object WalletConnect {
...
+                data class Text(val value: String) : Message()
+                class Bytes(val value: ByteArray) : Message()
+            }
+
+            data class ShutdownReason(val code: Int, val reason: String) : Model()
```
TalhaAli00, line 306: Relay as parent

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnect.kt
```diff
@@ -151,6 +152,159 @@ object WalletConnect {
             val chainId: String?,
             val params: String,
         ) : Model()
+
+        sealed class Relay : Model() {
+            sealed class DTO : Relay() {
```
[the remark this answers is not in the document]
Elyniss: How about: Payload, API, Data, Procedure, Action, Method, Call?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/WalletConnect.kt
```diff
@@ -151,6 +152,159 @@ object WalletConnect {
...
+
+                data class Error(
+                    val code: Long,
+                    val message: String,
+                ) {
+                    val errorMessage: String = "Error code: $code; Error message: $message"
+                }
+            }
+
+            sealed class Event : Model() {
```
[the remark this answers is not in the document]
Elyniss: Good catch! Thanks.

TalhaAli00 (approved):

PR #165, opened 2022-05-24
Renamed WalletConnect object to Sign

Elyniss:
Renamed AuthClient object to SignClient


TalhaAli00 (approved):

wallet/src/main/kotlin/com/walletconnect/wallet/WalletSampleApplication.kt
```diff
@@ -29,18 +29,18 @@ class WalletSampleApplication : Application() {
 //        )
 
         // Sample of how to use a URI to initialize the WalletConnect SDK
```
jakobuid, line 31:
```suggestion
        // Sample of how to use a URI to initialize the WalletConnect Sign SDK
```

jakobuid (approved):

PR #184, opened 2022-06-01
Refactor SignClient Singleton to enable multiple internal instances

Elyniss:


TalhaAli00 (approved):

jakobuid (approved):
I like it. Good job.
Just one thought. 
Once we have alpha for Chat SDK we should extract RelayClient to core to allow using the same instance of web-socket connection in both protocols. It should a default behaviour.

TalhaAli00:
That makes sense. I also think RelayClient as it is is going to change with our introduction of policies

TalhaAli00 (commented):
One thing we need to test is if the SignProtocol is accessible from Java files. Teams like Argent are still using Java

Elyniss:
@TalhaAli00 
> One thing we need to test is if the SignProtocol is accessible from Java files. Teams like Argent are still using Java

I created a function to test it: 

```java
public class ViewModelInJava extends ViewModel {

    public void someFunc(){
        List<Sign.Model.Session> listOfSettledSessions = SignClient.INSTANCE.getListOfSettledSessions(); // Using SignClient
        SignProtocol signProtocol = new SignProtocol(); // Unable to instantiate SignProtocol due to internal modifier
    }
}
```

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/client/SignClient.kt
```diff
@@ -1,281 +1,7 @@
...
-        fun onConnectionStateChange(state: Sign.Model.ConnectionState)
-    }
-
-    @Throws(IllegalStateException::class)
-    private fun checkEngineInitialization() {
-        check(::engineInteractor.isInitialized) {
-            "WalletConnectClient needs to be initialized first using the initialize function"
-        }
-    }
+object SignClient : SignInterface by SignProtocol.instance {
```
TalhaAli00, line 3: @Elyniss add @JvmStatic to this. Then Java users don't have to use `SignClient.INSTANCE`

PR #187, opened 2022-06-07
Support/add cha cha poly support below sdk 28

jakobuid:
Support ChaCha20Poly1304 on devices below API 28.

### DIFF
--- a/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/relay/data/codec/ChaChaPolyCodec.kt
+++ b/walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/relay/data/codec/ChaChaPolyCodec.kt
@@ -17,7 +17,7 @@ internal class ChaChaPolyCodec : Codec {
         val nonceBytes = randomBytes(NONCE_SIZE)
         val cha20Poly1305 = ChaCha20Poly1305()
         val params = ParametersWithIV(KeyParameter(key.keyAsHex.hexToBytes()), nonceBytes)
-        cha20Poly1305.init(true, params)
+        cha20Poly1305.init(true, params) //note: in the debugging mode code throws InvalidArgumentException but it doesn't affects the final method execution
         val cipherText = ByteArray(cha20Poly1305.getOutputSize(input.size))
         val outputSize = cha20Poly1305.processBytes(input, 0, input.size, cipherText, 0)
         cha20Poly1305.doFinal(cipherText, outputSize)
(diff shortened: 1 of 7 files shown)


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/relay/data/codec/ChaChaPolyCodec.kt
```diff
@@ -38,18 +38,17 @@ internal class ChaChaPolyCodec : Codec {
         byteBuffer.get(nonce)
         byteBuffer.get(encryptedText)
 
-        val ivParameterSpec: AlgorithmParameterSpec = IvParameterSpec(nonce)
-        val keySpec = SecretKeySpec(key.keyAsHex.hexToBytes(), CHA_CHA_20)
-        val cipher = Cipher.getInstance(CHA_CHA_POLY)
-        cipher.init(Cipher.DECRYPT_MODE, keySpec, ivParameterSpec)
-        val message = cipher.doFinal(encryptedText)
+        val cha20Poly1305 = ChaCha20Poly1305()
```
TalhaAli00, line 41: should we make this a private global variable inside this class since it's used in both encrypt and decrypt

jakobuid:
> @jakobuid One more thing, should we also change minSdk value in gradle script?

It has not changed since quite long time. Currently we aim minSdk to 24. Do you want to go lower? If yes, we should have a very good reason.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/relay/data/codec/ChaChaPolyCodec.kt
jakobuid: ya that would make sense. i'm going to check that, thanks

TalhaAli00 (commented):
Some of the biggest wallets support SDK 23 such as Trust Wallet, MetaMask and Alpha Wallet. Gnosis Safe Wallet goes all the way to 21. These are some of our biggest supporters so we should really try to support their wallets without breaking anything

jakobuid:
> Some of the biggest wallets support SDK 23 such as Trust Wallet, MetaMask and Alpha Wallet. Gnosis Safe Wallet goes all the way to 21. These are some of our biggest supporters so we should really try to support their wallets without breaking anything

Totally agree on that.
- TrustWallet - **Kotlin/Swift SDK** - minSDK 23
- MetaMast - React-Native SDK - minSDK 23
- Alpha Wallet - **Kotlin/Swift SDK** - minSDK 23
- Gnosis Safe - **Kotlin/Swift SDK** - minSDK 21

From what I found BouncyCastle is able to support devices even on Android 3.0. 
I suggest aiming 23, going all the way to 21 might be risky and it's only 0.3% of market.
What are your thoughts guys?

jakobuid:
After downgrading to sdk 23 there is an issue with NetworkState `registerDefaultNetworkCallback`. Issue: https://github.com/WalletConnect/WalletConnectKotlinV2/issues/191

PR #188, opened 2022-06-08
Updated error codes

Elyniss:


walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/Validator.kt
```diff
@@ -13,46 +27,63 @@ import java.net.URISyntaxException
...
+            !areProposalNamespacesKeysProperlyFormatted(namespaces) ->
+                onError(PeerError.UnsupportedNamespaceKey)
+            !areChainsNotEmpty(namespaces) ->
+                onError(PeerError.UnsupportedChains(NAMESPACE_CHAINS_MISSING_MESSAGE))
+            !areChainIdsValid(namespaces) ->
+                onError(PeerError.UnsupportedChains(NAMESPACE_CHAINS_CAIP_2_MESSAGE))
+            !areChainsInMatchingNamespace(namespaces) ->
+                onError(PeerError.UnsupportedChains(NAMESPACE_CHAINS_WRONG_NAMESPACE_MESSAGE))
+            !areExtensionChainsNotEmpty(namespaces) ->
+                onError(PeerError.UnsupportedChains(NAMESPACE_EXTENSION_CHAINS_MISSING_MESSAGE))
```
TalhaAli00, line 41: This looked better when on single lines :) just a comment

walletconnectv2/src/debug/kotlin/com.walletconnect.walletconnectv2/di/StorageModule.kt
```diff
@@ -9,7 +9,9 @@ import com.squareup.sqldelight.android.AndroidSqliteDriver
 import com.squareup.sqldelight.db.SqlDriver
 import com.walletconnect.walletconnectv2.Database
 import com.walletconnect.walletconnectv2.core.model.type.enums.MetaDataType
-import com.walletconnect.walletconnectv2.storage.data.dao.*
+import com.walletconnect.walletconnectv2.storage.data.dao.MetaDataDao
+import com.walletconnect.walletconnectv2.storage.data.dao.NamespaceDao
+import com.walletconnect.walletconnectv2.storage.data.dao.NamespaceExtensionsDao
```
jakobuid, line 14: Why this changed?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/core/exceptions/peer/PeerError.kt
```diff
@@ -87,18 +84,29 @@ sealed class PeerError {
         override val code: Int = 1301
     }
 
-    class UserDisconnected : PeerError() {
-        override val message: String = "User disconnected"
+    object UserDisconnected : PeerError() {
```
jakobuid, line 87: This one is actually not a PeerError, it's more like only an information that e.g. session has been disconnected. I would rather create a Reason sealed class. What do you think?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -68,8 +69,8 @@ internal class EngineInteractor(
             proposedRelays: List<EngineDO.RelayProtocolOptions>?,
             proposedSequence: EngineDO.ProposedSequence,
         ) {
-            Validator.validateProposalNamespace(namespaces.toNamespacesVOProposal()) { errorMessage ->
-                throw WalletConnectException.InvalidNamespaceException(errorMessage)
+            Validator.validateProposalNamespace(namespaces.toNamespacesVOProposal()) { peerError ->
+                throw WalletConnectException.InvalidNamespaceException(peerError.message)
```
jakobuid, line 73: PeerErrors were designed to send error between peers and are agreed among all platforms.
WalletConnectExceptions were design to let the end user know about an error too but we have here more flexibility about the message and what we want to throw.
I'd rather avoid mixing those two things.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/Validator.kt
```diff
@@ -13,46 +27,63 @@ import java.net.URISyntaxException
 internal object Validator {
 
     @JvmSynthetic
-    internal inline fun validateProposalNamespace(namespaces: Map<String, NamespaceVO.Proposal>, onNamespaceError: (String) -> Unit) {
+    internal inline fun validateProposalNamespace(namespaces: Map<String, NamespaceVO.Proposal>, onError: (PeerError) -> Unit) {
```
jakobuid, line 30: As above. I think it's better to keep Validator unaware of PeerError class. It's not a big deal, but we have more flexibility then.
Also, validation happens in many places in our code, however not always returned error is related to PeerError.
Does it make sense?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/Validator.kt
```diff
@@ -13,46 +27,63 @@ import java.net.URISyntaxException
...
-            !areExtensionAccountsNotEmpty(sessionNamespaces) -> onNamespaceError(NAMESPACE_EXTENSION_MISSING_ACCOUNTS_MESSAGE)
+                onError(PeerError.UserRejectedChains(NAMESPACE_ACCOUNTS_MISSING_FOR_CHAINS_MESSAGE))
+            !areAllMethodsApproved(sessionNamespaces, proposalNamespaces) ->
+                onError(PeerError.UserRejectedMethods)
+            !areAllEventsApproved(sessionNamespaces, proposalNamespaces) ->
+                onError(PeerError.UserRejectedEvents)
+            !areAccountsInMatchingNamespace(sessionNamespaces) ->
+                onError(PeerError.UserRejectedChains(NAMESPACE_ACCOUNTS_WRONG_NAMESPACE_MESSAGE))
+            !areExtensionAccountsNotEmpty(sessionNamespaces) ->
+                onError(PeerError.UserRejectedChains(NAMESPACE_EXTENSION_ACCOUNTS_MISSING_MESSAGE))
```
jakobuid, line 67: Formatting.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/Validator.kt
```diff
@@ -13,46 +27,63 @@ import java.net.URISyntaxException
...
+            !areSessionNamespacesKeysProperlyFormatted(sessionNamespaces) ->
+                onError(PeerError.UnsupportedNamespaceKey)
+            !areAccountsNotEmpty(sessionNamespaces) ->
+                onError(PeerError.UserRejectedChains(NAMESPACE_ACCOUNTS_MISSING_MESSAGE))
+            !areAccountIdsValid(sessionNamespaces) ->
+                onError(PeerError.UserRejectedChains(NAMESPACE_ACCOUNTS_CAIP_10_MESSAGE))
+            !areAccountsInMatchingNamespace(sessionNamespaces) ->
+                onError(PeerError.UserRejectedChains(NAMESPACE_ACCOUNTS_WRONG_NAMESPACE_MESSAGE))
+            !areExtensionAccountsNotEmpty(sessionNamespaces) ->
+                onError(PeerError.UserRejectedChains(NAMESPACE_EXTENSION_ACCOUNTS_MISSING_MESSAGE))
```
jakobuid, line 86: Formatting.

jakobuid (changes requested):

walletconnectv2/src/debug/kotlin/com.walletconnect.walletconnectv2/di/StorageModule.kt
Elyniss: this was an effect of `Optimize Imports`

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/core/exceptions/peer/PeerError.kt
Elyniss: I actually had it that way just before pushing to remote. It was called `PeerReason` and had only this `UserDisconnected`. I could reverse that. What that be better?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
Elyniss: I see your point. I will propose another solution

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/Validator.kt
Elyniss: I did single lines and then run auto formatting. Check now, please

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/core/exceptions/peer/PeerError.kt
```diff
@@ -58,22 +58,19 @@ internal sealed class PeerError {
```
jakobuid, line 58: `import com.walletconnect.walletconnectv2.core.exceptions.*` is not used.

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/model/mapper/ErrorMapper.kt
```diff
@@ -0,0 +1,19 @@
+package com.walletconnect.walletconnectv2.engine.model.mapper
+
```
jakobuid, line 2: Any reason not having it inside EngineMapper class?

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
```diff
@@ -4,6 +4,7 @@ import android.database.sqlite.SQLiteException
 import com.walletconnect.walletconnectv2.core.exceptions.*
```
jakobuid, line 4: On line 190 and 223 in EngineInteractor change peerError to error

jakobuid (changes requested):

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/model/mapper/ErrorMapper.kt
Elyniss: Sounds right. Moved it there

walletconnectv2/src/main/kotlin/com/walletconnect/walletconnectv2/engine/domain/EngineInteractor.kt
Elyniss: Thanks for noticing, found one more at 576 as well

jakobuid (approved):

PR #214, opened 2022-06-17
Added onError callback to delegates

TalhaAli00:


signSDK/sign/src/main/kotlin/com/walletconnect/sign/engine/model/EngineDO.kt
```diff
@@ -1,16 +1,20 @@
...
 
 import com.squareup.moshi.JsonClass
+import com.walletconnect.sign.core.exceptions.client.WalletConnectException
 import com.walletconnect.sign.core.model.type.EngineEvent
 import com.walletconnect.sign.core.model.type.Sequence
 import com.walletconnect.sign.core.model.vo.ExpiryVO
 import com.walletconnect.sign.core.model.vo.SecretKey
 import com.walletconnect.sign.core.model.vo.TopicVO
 import com.walletconnect.sign.core.model.vo.clientsync.common.RelayProtocolOptionsVO
+import java.lang.Exception
```
jakobuid, line 11: Not used

signSDK/sign/src/main/kotlin/com/walletconnect/sign/relay/domain/RelayerInteractor.kt
```diff
@@ -233,4 +236,11 @@ internal class RelayerInteractor(
     }
 
     private fun shouldPrompt(method: String): Boolean = method == JsonRpcMethod.WC_SESSION_REQUEST || method == JsonRpcMethod.WC_SESSION_PROPOSE
+
+    private fun handleError(errorMessage: String) {
```
jakobuid, line 240: Any reason of not emitting internal error on lines 205, 198, 166, 151, 141, 132 and 127?

jakobuid (changes requested):

signSDK/sign/src/main/kotlin/com/walletconnect/sign/relay/domain/RelayerInteractor.kt
TalhaAli00: Good catch on 198 and 205. The rest of the lines are already handled in their respective onFailure callbacks. There's no reason to emit the same error twice. This internal error should be only for things that the developer does not already receive through a normal workflow

signSDK/sign/src/main/kotlin/com/walletconnect/sign/engine/model/EngineDO.kt
TalhaAli00: removed

signSDK/sign/src/main/kotlin/com/walletconnect/sign/relay/domain/RelayerInteractor.kt
TalhaAli00: added other handleError calls

Elyniss (approved):

signSDK/sign/src/main/kotlin/com/walletconnect/sign/relay/domain/RelayerInteractor.kt
TalhaAli00: imo, the internal validations should be fast failures. If a validation fails, I don't think a developer should wait to catch the error in the delegate. We're basically doing a catch and if a developer is not looking closely as the `onError` function, then the validation error could be missed.

As for the `respondWithError`, I'm on the fence. Since the SDK automatically responds to the peer with an error, does a developer need to know the peer sent a back request?

TalhaAli00: @WalletConnect/kotlin-team

Elyniss: > As for the respondWithError, I'm on the fence. Since the SDK automatically responds to the peer with an error, does a developer need to know the peer sent a back request?

I don't think so

PR #215, opened 2022-06-17
Change waku to iridium

jakobuid:


signSDK/sign/src/main/kotlin/com/walletconnect/sign/network/model/RelayDTO.kt
```diff
@@ -23,7 +23,7 @@ internal sealed class RelayDTO {
             @Json(name = "jsonrpc")
             override val jsonrpc: String = "2.0",
             @Json(name = "method")
-            val method: String = "waku_publish",
+            val method: String = "iridium_publish",
```
TalhaAli00, line 26: make these strings constants that can re-used in Sign and RelayDTO

Elyniss (changes requested):
Can we extract all those methods to some file?

Elyniss (approved):

TalhaAli00 (approved):

PR #216, opened 2022-06-17
Feature/support android sdk 23

jakobuid:
- This PR adds the NetworkRequest callback that is set with wifi and cellular network types.
- It enables sdk to detect the connection change on those two network types.
- It is supported since API 21


signSDK/sign/src/main/kotlin/com/walletconnect/sign/util/NetworkState.kt
```diff
@@ -3,22 +3,59 @@ package com.walletconnect.sign.util
...
 internal class NetworkState(context: Context) {
 
-    private val _isAvailable = MutableStateFlow(false)
-    val isAvailable: StateFlow<Boolean> = _isAvailable.asStateFlow()
+    private val connectivityManager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+    private val _isWiFiAvailable = MutableStateFlow(false)
+    private val _isCellularAvailable = MutableStateFlow(false)
+
+    val isAvailable: StateFlow<Boolean> = combine(_isWiFiAvailable, _isCellularAvailable)
+    { wifi, cellular -> !(!wifi && !cellular) }.stateIn(scope, SharingStarted.Eagerly, false)
```
TalhaAli00, line 18: the boolean logic can be simplified to `wifi || cellular`

TalhaAli00: The default should be true

Elyniss: Why defaulting to true? false seems more fault-proof for me

signSDK/dapp/build.gradle.kts
```diff
@@ -9,7 +9,7 @@ android {
 
     defaultConfig {
         applicationId = "com.walletconnect.dapp"
-        minSdk = 24
+        minSdk = 23
```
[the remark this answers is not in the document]
jakobuid: Ya, makes sens

signSDK/sign/src/main/kotlin/com/walletconnect/sign/util/NetworkState.kt
jakobuid: ya, the condition could be simplified

I agree with @Elyniss because when sdk is initialized and connection is detected the state flow changes value from false to true.

TalhaAli00: but I'm ok with either one

TalhaAli00 (approved):

signSDK/sign/src/main/kotlin/com/walletconnect/sign/util/NetworkState.kt
jakobuid: Good point. I'll test such scenario once we have relay situation stabilised.

TalhaAli00 (approved):

Elyniss (approved):

PR #218, opened 2022-06-20
Chat SDK setup

Elyniss:


chatSDK/chat/src/androidTest/kotlin/com/walletconnect/chat/ExampleInstrumentedTest.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+@RunWith(AndroidJUnit4::class)
+class ExampleInstrumentedTest {
+    @Test
+    fun useAppContext() {
+        // Context of the app under test.
+        val appContext = InstrumentationRegistry.getInstrumentation().targetContext
+        assertEquals("com.walletconnect.chat.test", appContext.packageName)
+    }
+}
```
jakobuid, line 24: If not needed lets remove?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatInterface.kt
```diff
@@ -0,0 +1,40 @@
...
+    fun reject(reject: Chat.Params.Reject, onError: (Chat.Model.Error) -> Unit)
+    fun message(message: Chat.Params.Message, onError: (Chat.Model.Error) -> Unit)
+    fun ping(ping: Chat.Params.Ping, onError: (Chat.Model.Error) -> Unit)
+    fun leave(leave: Chat.Params.Leave, onError: (Chat.Model.Error) -> Unit)
+    fun addContact(addContact: Chat.Params.AddContact, onError: (Chat.Model.Error) -> Unit)
+    fun getInvites(getInvites: Chat.Params.GetInvites): Map<String, Chat.Model.Invite>
+    fun getThreads(getThreads: Chat.Params.GetThreads): Map<String, Chat.Model.Thread>
+    fun getMessages(getMessages: Chat.Params.GetMessages): List<Chat.Model.Message>
+/***
+ TODO: Relay should be extracted to core along with websocket interface
```
jakobuid, line 26: 👍

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatProtocol.kt
```diff
@@ -0,0 +1,99 @@
+package com.walletconnect.chat.client
+
+import com.walletconnect.chat.engine.domain.EngineInteractor
+import org.koin.core.KoinApplication
+
+internal class ChatProtocol : ChatInterface {
+    private val wcKoinApp: KoinApplication = KoinApplication.init()
+    private lateinit var engineInteractor: EngineInteractor
```
jakobuid, line 8: Maybe we could change `EngineInteractor` to `ChatEngine` and in sign module to `SignEngine`?
What do you guys think? @WalletConnect/kotlin-team

jakobuid (changes requested):

jakobuid (approved):

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatProtocol.kt
Elyniss: I agree.

TalhaAli00: I like this idea

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatInterface.kt
TalhaAli00: For now, can we duplicate the code into Chat? I'm curious if Relay will change over time with the Chat SDK work

Elyniss: If we just copy code and create another instance we would have two websockets connections, right? For MVP I'm not against it but later on we should at investigate the solution

jakobuid: For MVP it is good enough

PR #223, opened 2022-06-21
KeyServer API

Elyniss:


signSDK/samples_common/src/main/kotlin/com/walletconnect/sample_common/Constants.kt
```diff
@@ -1,4 +1,4 @@
 package com.walletconnect.sample_common
 
 const val WALLET_CONNECT_PROD_RELAY_URL = "relay.walletconnect.com"
-const val PROJECT_ID = ""
\ No newline at end of file
+const val PROJECT_ID = BuildConfig.PROJECT_ID
```
Elyniss, line 4: I've added ENV VAR:
`export WC_CLOUD_PROJECT_ID=2ee94aca5d98e6c05c38bce02bee952a`

With this we won't have to worry about having it inside the repo.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/di/EngineModule.kt
```diff
@@ -0,0 +1,10 @@
+package com.walletconnect.chat.di
```
TalhaAli00, line 1: add JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/di/KeyServerModule.kt
```diff
@@ -0,0 +1,50 @@
+package com.walletconnect.chat.di
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/di/KeyServerModule.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.walletconnect.chat.discovery.keyserver.domain.use_case.ResolveAccountUseCase
+import okhttp3.OkHttpClient
+import org.koin.dsl.module
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+import java.util.concurrent.TimeUnit
+
+@JvmSynthetic
+internal fun keyServerModule(keyServerUrl: String) = module {
+    val TIMEOUT_TIME = 5000L
```
TalhaAli00, line 17: can we make this private?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/di/KeyServerModule.kt
```diff
@@ -0,0 +1,50 @@
...
+    single {
+        get<Retrofit>().create(KeyServerService::class.java)
+    }
+
+    single<KeyServerClient> { KeyServerClientImpl(get()) }
+
+    single {
+        KeyServerInteractor(
+            RegisterAccountUseCase(get()),
+            ResolveAccountUseCase(get())
```
TalhaAli00, line 46: I think it would be better to create the different UseCases as separate singletons then inject the use case into Interactor

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/data/client/KeyServerClient.kt
```diff
@@ -0,0 +1,8 @@
+package com.walletconnect.chat.discovery.keyserver.data.client
+
+import com.walletconnect.chat.discovery.keyserver.model.KeyServerDTO
+
+internal interface KeyServerClient {
+    suspend fun register(account: KeyServerDTO.Account)
+    suspend fun resolve(account: String): KeyServerDTO.Account
+}
```
TalhaAli00, line 8: any reason why this needs to be an interface? We only have one impl, can't this be a concrete class?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/KeyServerInteractor.kt
```diff
@@ -0,0 +1,14 @@
+package com.walletconnect.chat.discovery.keyserver.domain
+
+import com.walletconnect.chat.engine.model.EngineDO
+import com.walletconnect.chat.discovery.keyserver.domain.use_case.RegisterAccountUseCase
+import com.walletconnect.chat.discovery.keyserver.domain.use_case.ResolveAccountUseCase
+
+internal class KeyServerInteractor(
```
TalhaAli00, line 7: The interactor is just a passthrough class here, I think we can just inject the usecases into the ChatEngine

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/RegisterAccountUseCase.kt
```diff
@@ -0,0 +1,22 @@
...
+import com.walletconnect.chat.discovery.keyserver.data.client.KeyServerClient
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
+
+internal class RegisterAccountUseCase(
+    private val client: KeyServerClient,
+) {
+    operator fun invoke(account: EngineDO.Account): Flow<Result<Unit>> = flow {
```
TalhaAli00, line 14: does this have to return a flow?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
```diff
@@ -1,4 +1,43 @@
 package com.walletconnect.chat.engine.domain
```
TalhaAli00, line 1: add JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/model/KeyServerDTO.kt
```diff
@@ -0,0 +1,12 @@
+package com.walletconnect.chat.discovery.keyserver.model
```
TalhaAli00, line 1: add JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/ResolveAccountUseCase.kt
```diff
@@ -0,0 +1,22 @@
+package com.walletconnect.chat.discovery.keyserver.domain.use_case
```
TalhaAli00, line 1: add JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/RegisterAccountUseCase.kt
```diff
@@ -0,0 +1,22 @@
+package com.walletconnect.chat.discovery.keyserver.domain.use_case
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/data/service/KeyServerService.kt
```diff
@@ -0,0 +1,15 @@
+package com.walletconnect.chat.discovery.keyserver.data.service
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/data/client/KeyServerClientImpl.kt
```diff
@@ -0,0 +1,11 @@
+package com.walletconnect.chat.discovery.keyserver.data.client
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/data/client/KeyServerClient.kt
```diff
@@ -0,0 +1,8 @@
+package com.walletconnect.chat.discovery.keyserver.data.client
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/model/EngineDO.kt
```diff
@@ -1,4 +1,50 @@
 package com.walletconnect.chat.engine.model
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/model/EngineMapper.kt
```diff
@@ -1,2 +1,24 @@
 package com.walletconnect.chat.engine.model
```
TalhaAli00, line 1: JvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/model/EngineMapper.kt
```diff
@@ -1,2 +1,24 @@
...
 
+import com.walletconnect.chat.client.Chat
+import com.walletconnect.chat.discovery.keyserver.model.KeyServerDTO
+
+// TODO: Organize Mappers - Where should mapping methods between client and engine be (ClientMapper.kt or EngineMapper.kt)?
```
TalhaAli00, line 6: The pattern we decided on using for the Sign SDK was to only expose an object with the ending `VO` (which means Value Object) between layers (data -> domain). So the repository should convert all DTO objects to a VO object and inside the EngineMapper, you map the VO object to a DO object which is used inside of the Engine.

The benefit is that we can expose only the data we need and remove things like IDs or anything else the DTO will need or DB objects.

chatSDK/chatsample/src/main/kotlin/com/walletconnect/chatsample/ui/host/ChatSampleActivity.kt
```diff
@@ -0,0 +1,69 @@
...
+        viewModel.events
+            .flowWithLifecycle(lifecycle)
+            .onEach { event ->
+                when (event) {
+                    is ChatSampleEvents.OnResolve -> onResolve(event.publicKey)
+                    is ChatSampleEvents.OnRegister -> onRegister(event.publicKey)
+                    else -> Unit
+                }
+            }
+            .launchIn(lifecycleScope)
```
TalhaAli00, line 41: nice

chatSDK/chatsample/src/main/kotlin/com/walletconnect/chatsample/ui/host/ChatSampleViewModel.kt
```diff
@@ -0,0 +1,51 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.shareIn
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class ChatSampleViewModel : ViewModel() {
+    val events = ChatDelegate.wcEventModels.map { wcEvent ->
+        when (wcEvent) {
+            is Chat.Model.Events.OnResolve -> ChatSampleEvents.OnResolve(wcEvent.publicKey)
+            is Chat.Model.Events.OnRegister -> ChatSampleEvents.OnRegister(wcEvent.publicKey)
```
TalhaAli00, line 20: are these events only supposed to fire once?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/di/KeyServerModule.kt
Elyniss: This is `fun`. I'm not sure what you mean

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/KeyServerInteractor.kt
Elyniss: Yes, we can. I was just sugarcoating the architecture. My idea was to aggregate all use-cases into one interactor so that engine won't have multiple use-cases in the constructor. 

Also, this class could later be renamed to `DiscoverabilityInteractor` where we could fit use-cases for resolving ens public keys. 

Given that, what do you think? @TalhaAli00

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/RegisterAccountUseCase.kt
Elyniss: Nice catch, it was a residue from my tinkering with custom `Result` that supports `Loading` state, but I abandoned the idea. I will remove Flow and stick to Result

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/model/EngineMapper.kt
Elyniss: So to sum up this JvmSynth annotation. We need to add this to every file that has something internal inside it, right?

chatSDK/chatsample/build.gradle.kts
```diff
@@ -23,19 +23,26 @@ android {
...
+    buildFeatures {
+        viewBinding = true
     }
 }
 
 dependencies {
+    implementation(project(":chat"))
+    implementation(project(":samples_common"))
 
+    implementation("com.andrognito.flashbar:flashbar:1.0.3")
```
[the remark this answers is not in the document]
Elyniss: Moved to snackbar

chatSDK/chatsample/src/main/kotlin/com/walletconnect/chatsample/ui/host/ChatSampleViewModel.kt
Elyniss: Yeah, nice catch. I've added a clear cache method to `ChatDelegate`

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/model/EngineMapper.kt
Elyniss: I think I understand it now. Take a look at introduced with new commit VO classes and mappings

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/di/KeyServerModule.kt
jakobuid: I guess it's about making `TIMEOUT_TIME` private, but it doesn't make sense because it is scoped to module block.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/KeyServerInteractor.kt
jakobuid: I like your idea @Elyniss however I'm against doing optimization when it doesn't bring any value to the project architecture.

I'd rather do it when it's really needed, as you said for instance when we'd have to add resolving ens public keys feature.

I think now is just too early.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/RegisterAccountUseCase.kt
```diff
@@ -0,0 +1,19 @@
...
+internal class RegisterAccountUseCase(
+    private val client: KeyServerClient,
+) {
+    suspend operator fun invoke(account: AccountVO): Result<Unit> = try {
+        client.register(account.toDTOAccount())
+        Result.success(Unit)
+    } catch (e: Exception) {
+        Result.failure(e)
+    }
+
```
jakobuid, line 18: Empty space

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/ResolveAccountUseCase.kt
```diff
@@ -0,0 +1,19 @@
...
+
+internal class ResolveAccountUseCase(
+    private val client: KeyServerClient,
+) {
+    suspend operator fun invoke(account: String): Result<AccountVO> = try {
+        val accountVO = client.resolve(account).toVOAccount()
+        Result.success(accountVO)
+    } catch (e: Exception) {
+        Result.failure(e)
+
```
jakobuid, line 17: Empty space

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/use_case/RegisterAccountUseCase.kt
```diff
@@ -0,0 +1,19 @@
...
+package com.walletconnect.chat.discovery.keyserver.domain.use_case
+
+import com.walletconnect.chat.core.model.vo.AccountVO
+import com.walletconnect.chat.discovery.keyserver.data.client.KeyServerClient
+import com.walletconnect.chat.discovery.keyserver.model.toDTOAccount
+
+internal class RegisterAccountUseCase(
+    private val client: KeyServerClient,
+) {
+    suspend operator fun invoke(account: AccountVO): Result<Unit> = try {
```
jakobuid, line 12: Here inside `AccountVO` you have `account` and `publicKey` but in the `ResolveAccountUseCase` you pass account of type string.

It's a bit confusing because we have two definitions of account then.

One is account address with pubKey, the other is just account address.
Maybe we could just jus AccountVO with optional pubKey?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
```diff
@@ -1,4 +1,42 @@
...
+internal class ChatEngine(
+    private val keyServerInteractor: KeyServerInteractor,
+) {
+    private val _events: MutableSharedFlow<EventsVO> = MutableSharedFlow()
+    val events: SharedFlow<EventsVO> = _events.asSharedFlow()
+
+    internal fun resolveAccount(account: String, onFailure: (Throwable) -> Unit) {
+        scope.launch {
+            keyServerInteractor.resolveAccount(account).fold(
+                onSuccess = { account -> _events.emit(EngineDO.Events.OnResolve(account.publicKey).toVOOnResolve()) },
```
jakobuid, line 28: Do we really need to emit an event here?
I think if we use the onSuccess callback as you do with error it should be all good. Or even `onResult(Result<OnResolve>)`

This is single http request where success or failure is received almost immediately.

It is done in this way in Sing because we publish a message over wss, where such message has to be handled by the other peer before receiving ack.

What do you think?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
```diff
@@ -1,4 +1,42 @@
...
+                onFailure = { error -> withContext(Dispatchers.Main) { onFailure(error) } }
+            )
+        }
+    }
+
+    internal fun registerAccount(account: String, publicKey: String, onFailure: (Throwable) -> Unit) {
+        scope.launch {
+            keyServerInteractor.registerAccount(AccountVO(account, publicKey)).fold(
+                onSuccess = { _events.emit(EngineDO.Events.OnRegister(publicKey).toVOOnRegister()) },
```
jakobuid, line 37: Same

chatSDK/chatsample/src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,25 @@
...
+    <application
+        android:name=".ChatSampleApplication"
+        android:allowBackup="true"
+        android:icon="@drawable/ic_walletconnect_circle_blue"
+        android:label="@string/app_name"
+        android:roundIcon="@drawable/ic_walletconnect_circle_blue"
+        android:supportsRtl="true"
+        android:theme="@style/Theme.WalletConnect"
+        android:usesCleartextTraffic="true">
+<!--TODO: Remove android:usesCleartextTraffic="true"-->
```
jakobuid, line 13: Why not just remove it? :)

chatSDK/chatsample/src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,25 @@
...
+        <activity
+            android:name=".ui.host.ChatSampleActivity"
+            android:exported="true">
```
jakobuid, line 16: Do we wanna have only portret mode?

jakobuid (changes requested):

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/Chat.kt
```diff
@@ -19,30 +19,42 @@ object Chat {
...
+            data class OnRegister(val publicKey: String) : Events() // TODO: Not sure if publicKey is necessary in here and in specs
         }
     }
 
     sealed class Params {
-        data class Init(val application: Application) : Params()
+        data class Init(val application: Application, val keyServerUrl: String) : Params()
 
-        data class Register(val account: String, val private: Boolean? = false) : Params()
+        data class Register(val account: String, val publicKey: String, val private: Boolean? = false) : Params()
```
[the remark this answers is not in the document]
Elyniss: This key is always a public key. From what I understood from the specs & discussions, after MVP the `private` flag will tell whether we should register a given account in our KeyServer.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
jakobuid: Plus we don't need to scope after we receive a result. So it's better to use supervisorScope and cancel it whenever we success or error.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/copiedFromSign/core/scope/WalletConnectScope.kt
```diff
@@ -0,0 +1,13 @@
+@file:JvmSynthetic
+
+package com.walletconnect.chat.copiedFromSign.core.scope
```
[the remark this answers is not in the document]
Elyniss: Todo added, however, the idea behind naming module `copiedFromSign` is to indicate exactly what you propose.

chatSDK/chatsample/src/main/AndroidManifest.xml
Elyniss: Right now keyserver has no TLS cert

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/KeyServerInteractor.kt
Elyniss: I honestly believe it does bring value right now. 

However, if both of you agree it doesn't bring value I will gladly remove `KeyServerInteractor`

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
jakobuid: I think Engine should not be aware of any thread switching. If developers want to update an UI on onFailure, it should be in their side to remember about the handling it on the right thread.

It applies for both the onSuccess and the onError.

We shouldn't force any context switching.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/discovery/keyserver/domain/KeyServerInteractor.kt
Elyniss: Removed it

signSDK/sign/src/main/kotlin/com/walletconnect/sign/crypto/KeyManagementRepository.kt
```diff
@@ -17,4 +17,6 @@ internal interface KeyManagementRepository {
     fun generateKeyPair(): PublicKey
 
     fun removeKeys(tag: String)
+
+    // Needs to be added with Chat SDK
```
jakobuid, line 21: ?

jakobuid (changes requested):

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/copiedFromSign/crypto/KeyManagementRepository.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun getSymmetricKey(topic: TopicVO): SymmetricKey
+
+    fun generateSymmetricKeyFromKeyAgreement(self: PublicKey, peer: PublicKey): SymmetricKey
+    fun generateTopicFromKeyAgreement(self: PublicKey, peer: PublicKey): TopicVO
+    fun getKeyAgreement(topic: TopicVO): Pair<PublicKey, PublicKey>
+    fun generateKeyPair(): PublicKey
+
+    fun removeKeys(tag: String)
+
+    //Added with Chat SDK
```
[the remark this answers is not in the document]
Elyniss: It was only to explicitly say what was added in Chat SDK. Right now there are two files `KeyManagementRepository.kt` . I just added it to not be confused if whose two files become out of sync later on.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/Chat.kt
```diff
@@ -4,13 +4,24 @@ import android.app.Application
...
+        interface Resolve : Listeners {
+            fun onSuccess(publicKey: String)
+            fun onError(error: Model.Error)
+        }
+        interface Register : Listeners {
+            fun onSuccess(publicKey: String)
+            fun onError(error: Model.Error)
+        }
+    }
+
```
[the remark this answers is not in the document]
Elyniss: I don't like the fact that every Listener has `onSuccess` with a predefined String parameter. How about this solution

```suggestion
    sealed interface Listeners {
      fun onError(error: Model.Error)
      private interface PublicKeyOnSuccess {
        fun onSuccess(publicKey: String)
      } : Listeners
      
      interface Resolve : PublicKeyOnSuccess
      interface Register : PublicKeyOnSuccess
    }

```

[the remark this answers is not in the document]
jakobuid: Yea looks good!

signSDK/sign/src/main/kotlin/com/walletconnect/sign/crypto/KeyManagementRepository.kt
Elyniss: Removed

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/Chat.kt
```diff
@@ -19,28 +30,36 @@ object Chat {
         ) : Model()
 
         data class Thread(
-            // TODO: Define structure in specs
             val topic: String,
+            val selfAccount: String,
+            val peerAccount: String,
```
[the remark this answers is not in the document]
Elyniss: Good call. Added AccountId value class. Also added `fun isValid()` for SDK consumer to validate correctness

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/Chat.kt
```diff
@@ -1,16 +1,34 @@
...
+        value class AccountId(val value: String) {
+            fun isValid() = AccountIdVO(value).isValid()
+        }
+
         data class Invite(
-            val account: String,
+            val account: AccountId,
             val message: String,
-            val signature: String? = "", // TODO: Extract String.Empty to core module and use it here
+            val signature: String? = null,
```
TalhaAli00, line 31: should this be defaulted to null?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/Chat.kt
```diff
@@ -1,16 +1,34 @@
...
+        interface Register : PublicKeyOnSuccess
+    }
+
+
     sealed class Model {
         data class Error(val throwable: Throwable) : Model() // TODO: Should this be extracted to core for easier error handling?
 
+        @JvmInline
+        value class AccountId(val value: String) {
+            fun isValid() = AccountIdVO(value).isValid()
```
TalhaAli00, line 25: instead of wrapping value with `AccountIdVO`, can't you just run the validator logic here? We should try to not expose our internal object like VO classes

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatProtocol.kt
```diff
@@ -1,6 +1,15 @@
 package com.walletconnect.chat.client
```
TalhaAli00, line 1: add jvmsynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatProtocol.kt
```diff
@@ -14,20 +23,56 @@ internal class ChatProtocol : ChatInterface {
...
+                    is EventsVO.OnLeft -> TODO()
+                    is EventsVO.OnMessage -> TODO()
+                }
+            }
+        }
     }
 
     @Throws(IllegalStateException::class)
-    override fun register(register: Chat.Params.Register, onError: (Chat.Model.Error) -> Unit) {
+    override fun register(register: Chat.Params.Register, listener: Chat.Listeners.Register) {
```
TalhaAli00, line 56: is the reason for changing from lamdas to interfaces for Java support?

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/copiedFromSign/crypto/data/keystore/KeyChain.kt
```diff
@@ -0,0 +1,50 @@
...
+        sharedPreferences.edit().remove(tag).apply()
+    }
+
+    private fun concatKeys(keyA: Key, keyB: Key): String = (keyA.keyAsHex.hexToBytes() + keyB.keyAsHex.hexToBytes()).bytesToHex()
+
+    private fun splitKeys(concatKeys: String): Pair<String, String> {
+        val concatKeysByteArray = concatKeys.hexToBytes()
+        val privateKeyByteArray = concatKeysByteArray.sliceArray(0 until (concatKeysByteArray.size / 2))
+        val publicKeyByteArray = concatKeysByteArray.sliceArray((concatKeysByteArray.size / 2) until concatKeysByteArray.size)
+        return privateKeyByteArray.bytesToHex() to publicKeyByteArray.bytesToHex()
```
TalhaAli00, line 43: maybe make the return type `Pair<PrivateKey, PublicKey>` to make sure that we don't mix up the two keys

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/copiedFromSign/crypto/data/keystore/KeyChain.kt
```diff
@@ -0,0 +1,50 @@
...
+        return sharedPreferences.getString(tag, String.Empty) ?: String.Empty
+    }
+
+    override fun setKeys(tag: String, key1: Key, key2: Key) {
+        val keys = concatKeys(key1, key2)
+        sharedPreferences.edit().putString(tag, keys).apply()
+    }
+
+
+    override fun getKeys(tag: String): Pair<String, String> {
```
TalhaAli00, line 28: make sure that the usage of this function is consistent

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/Validator.kt
```diff
@@ -0,0 +1,21 @@
+package com.walletconnect.chat.engine.domain
```
TalhaAli00, line 1: jvmSynth annotation

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/engine/model/EngineMapper.kt
TalhaAli00: correct, any internal class or file that has internal functions. This will prevent the files and classes from being seen when using Java since internal is a Kotlin specific feature

TalhaAli00 (approved):

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/ChatProtocol.kt
Elyniss: The main reason was that callback should not be handled as the event from our SDK like i.e. sessionPropose.

chatSDK/chat/src/main/kotlin/com/walletconnect/chat/client/Chat.kt
Elyniss: I think that doing it this way is not exposing the VO class. I didn't want to have multiple definitions of validation for public and internal AccountId value class

PR #276, opened 2022-07-29
added suffix const and changed iridium to irn

TalhaAli00:


Elyniss (changes requested):
I like your approach, however you only applied prefix changes into Chat SDK. Could you also add it to Sign SDK?

Elyniss:
Should we also change the default `protocol` value in `RelayProtocolOptionsVO`?

Elyniss:
`RelayerDOJsonRpcResponseJsonRpcResultJsonAdapterTest` contains "iridium" as well

Elyniss:
I would also rename `IridiumParamsVO` to `IrnParamsVo` with all it's usages

Elyniss:
Also could we remove `Iridium` prefix from `IridiumJWTHeader` and `IridiumJWTPayload` classes

Elyniss (approved):
LGTM

jakobuid (approved):

PR #277, opened 2022-07-29
Updated publishing to use Maven Central instead of Jitpack

TalhaAli00:


buildSrc/src/main/kotlin/publish-module.gradle.kts
```diff
@@ -0,0 +1,76 @@
+//apply plugin: signing
```
TalhaAli00, line 1: I'm working on this class to make publishing "injectable" by just adding this file to all sdk modules

trastvolt:
feature/maven-central-update

jakobuid (approved):

PR #289, opened 2022-08-01
feat: automatically moves issues to board

arein:
If a team member creates an issue it goes straight to the board, else it needs to be labeled as "accepted".

# Testing
Tested in the monorepo and just copy/pasta here.

### DIFF
--- a/.github/workflows/intake.yml
+++ b/.github/workflows/intake.yml
@@ -1,0 +1,43 @@
+# This workflow moves issues to the Kotlin board
+# when they receive the "accepted" label
+# When WalletConnect Org members create issues they
+# are automatically "accepted".
+# Else they need to manually receive that label during intake.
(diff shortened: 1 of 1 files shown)


jakobuid:
In the case then someone else other than @WalletConnect/kotlin-team creates an issue, the accepted label must be added manually by one of the team members and only then such issue is visible on the project board. Is that correct? @arein

arein:
> In the case when someone else other than @WalletConnect/kotlin-team creates an issue, the accepted label must be added manually by one of the team members and only then such issue is visible on the project board. Is that correct? @arein

Correct. The idea is to standardize the intake process across all public repos. @hackyguru is driving this.

Elyniss (approved):

jakobuid (approved):

Issue #296, opened 2022-08-01
Incorrect JSONRPCRequest.params format


Azkae:
Some params received in `JSONRPCRequest.params` are not in the correct format:

**Expected**

For example on method `eth_signTransaction` (https://docs.walletconnect.com/2.0/advanced/rpc-reference/ethereum-rpc#eth_signtransaction), params are sent by the dApp in the following format: 
```
[
  {
    "from": "0x0f364Bac605ba88200ac0c740d5484E6634B59d6",
    "to": "0x0f364Bac605ba88200ac0c740d5484E6634B59d6",
    "data": "0x",
    "nonce": "0x0",
    "gasPrice": "0x4a817c800",
    "gasLimit": "0x5208",
    "value": "0x0"
  }
]
```

Note that `value`, `nonce`, `gas` and `gasPrice` are strings.

**Actual behavior**

But `JSONRPCRequest.params` is in the following format:

```
[
  {
    "from": "0x0f364Bac605ba88200ac0c740d5484E6634B59d6",
    "to": "0x0f364Bac605ba88200ac0c740d5484E6634B59d6",
    "data": "0x",
    "nonce": 0,
    "gasPrice": 20000000000,
    "gasLimit": 21000,
    "value": 0
  }
]
```

`nonce`, `gasPrice`, `gasLimit` and `value` are integers instead of strings, see the documentation here: https://docs.walletconnect.com/2.0/advanced/rpc-reference/ethereum-rpc#eth_signtransaction.

I have not tested it but I suspect this will also cause an issue if `value` exceed a Kotlin Long (2^64) since it can go up to 2^256.

**SDK Version**
 - Client: Kotlin
 - Version `2.0.0-rc.0`

TalhaAli00:
@Azkae RC.1 was released today with a fix for the serializing

PR #298, opened 2022-08-02
updated serialization logic for SessionRequestVOJsonAdapter

TalhaAli00:


signSDK/sign/src/test/kotlin/com/walletconnect/sign/core/adapters/SessionRequestVOJsonAdapterTest.kt
```diff
@@ -0,0 +1,200 @@
...
+        val params = """
+            {
+                "transactions":[
+                      {
+                        "nonce":21,
+                        "value":"10000000000000000",
+                        "receiver":"erd1c542gysqkmfqwwwu7tkuhz8kp9yxkg32gzsuvgmryrs74nhhltzqkna8na",
+                        "sender":"erd1c542gysqkmfqwwwu7tkuhz8kp9yxkg32gzsuvgmryrs74nhhltzqkna8na",
+                        "gasPrice":1000000000,
+                        "gasLimit":60000000,
```
jakobuid, line 37: Let's keep `gasLimit`, `gasPrice` and `value` fields as strings according to documentation: https://docs.walletconnect.com/json-rpc-api-methods/ethereum#example-parameters-2
and test against string values.

Also, we have a bug report: https://github.com/WalletConnect/WalletConnectKotlinV2/issues/296
I think this code will fix that, what's your opinion?

TalhaAli00: The serializer will work even after changing it to a string. I think the only thing I have left to check is a boolean. I'll update the test and then hopefully we can merge

jakobuid (approved):

Elyniss (approved):

PR #321, opened 2022-08-04
Feature/deep linking new flow

jakobuid:
New flow of mobile deep linking between Dapp and Wallet.
Added the redirect field with native fields that enables redirection from Wallet to Dapp after handling session request.


signSDK/sign/src/main/sqldelight/migrations/1.sqm
```diff
@@ -2,8 +2,6 @@ import com.walletconnect.sign.core.model.type.enums.MetaDataType;
 import kotlin.collections.List;
 
 -- PairingDao
-ALTER TABLE PairingDao RENAME TO PairingDaoBak;
```
Elyniss, line 5: I think we cannot remove this

signSDK/sign/src/main/sqldelight/migrations/1.sqm
```diff
@@ -17,8 +15,6 @@ CREATE TABLE IF NOT EXISTS PairingDao (
 DROP TABLE IF EXISTS PairingDaoBak;
 
 -- SessionDao
-ALTER TABLE SessionDao RENAME TO SessionDaoBak;
```
Elyniss, line 20: Same as above

signSDK/sign/src/main/sqldelight/migrations/1.sqm
```diff
@@ -34,8 +30,6 @@ CREATE TABLE IF NOT EXISTS  SessionDao(
 DROP TABLE IF EXISTS SessionDaoBak;
 
 -- MetaDataDao
-ALTER TABLE MetaDataDao RENAME TO MetaDataDaoBak;
```
Elyniss, line 37: Same as above

signSDK/sign/src/main/sqldelight/migrations/1.sqm
```diff
@@ -50,8 +44,6 @@ CREATE TABLE IF NOT EXISTS MetaDataDao(
 DROP TABLE IF EXISTS MetaDataDaoBak;
 
 -- JsonRpcHistoryDao
-ALTER TABLE JsonRpcHistoryDao RENAME TO JsonRpcHistoryDaoBak;
```
Elyniss, line 53: Same as above

Elyniss (changes requested):

signSDK/sign/src/main/sqldelight/migrations/1.sqm
Elyniss: @TalhaAli00 what's your take on this?

jakobuid: After calling `verifySqlDelightMigration` this line was causing test to fail

Elyniss (approved):

signSDK/dapp/src/main/kotlin/com/walletconnect/dapp/DappSampleApplication.kt
```diff
@@ -28,15 +28,18 @@ class DappSampleApplication : Application() {
 //        )
 
         // Sample of how to use a URI to initialize the WalletConnect SDK
+
+        val redirect = Sign.Model.Redirect(native = "kotlin-dapp-wc:/request")
```
TalhaAli00, line 32: 1. Does the redirect have to be a object? Since we're only using the native property, could we potentially just have a property in Init? I don't see when the Android app would use the universal property. Please explain
2. Prior to this PR, deeplinks were split into `pairing` and `request`. Do we not want to keep this pattern for dapps?

signSDK/dapp/src/main/kotlin/com/walletconnect/dapp/ui/selected_account/SelectedAccountViewModel.kt
```diff
@@ -75,7 +75,11 @@ class SelectedAccountViewModel : ViewModel() {
                 }
             }
 
-            val sessionRequestDeepLinkUri = "wc:/${requestParams.sessionTopic})}/request".toUri()
+            val sessionRequestDeepLinkUri =
+                SignClient.getListOfSettledSessions()
+                    .find { session -> session.topic == requestParams.sessionTopic }?.metaData?.redirect?.native?.toUri()
```
TalhaAli00, line 80: I think we talked about having a simplified properties/helper functions to get the uri instead of going 3 layers deep to get the info a developer wants. Do we want to do this work later?

signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
```diff
@@ -142,6 +152,12 @@ object Sign {
             val description: String,
             val url: String,
             val icons: List<String>,
+            val redirect: Redirect?,
+        ) : Model()
+
+        data class Redirect(
+            val native: String? = null,
+            val universal: String? = null,
```
TalhaAli00, line 160: follow up to the above comment, is it possible to remove this for Android?

signSDK/wallet/src/main/kotlin/com/walletconnect/wallet/WalletSampleApplication.kt
```diff
@@ -29,6 +29,8 @@ class WalletSampleApplication : Application() {
 //        )
 
         // Sample of how to use a URI to initialize the WalletConnect Sign SDK
+
+        val redirect = Sign.Model.Redirect(native = "kotlin-wallet-wc:/request")
```
TalhaAli00, line 33: similar comment as above. We had a separate URI for pairing vs requests. Do we not want this anymore?

TalhaAli00 (approved):

signSDK/dapp/src/main/kotlin/com/walletconnect/dapp/DappSampleApplication.kt
jakobuid: 1. For me it makes sense. In db we keep only `native` field. I can't see any objection of removing `universal` and keep a structure flat. Good call. 
2. Deeplinks where always split into `pairing` and `request`.  Yes ofc we want to keep this pattern for Dapps, because it allows wallet to redirect directly to a Dapp after approving or rejecting a signing request.

signSDK/dapp/src/main/kotlin/com/walletconnect/dapp/ui/selected_account/SelectedAccountViewModel.kt
jakobuid: Nah, I can add it to this PR

signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
jakobuid: Yes

signSDK/wallet/src/main/kotlin/com/walletconnect/wallet/WalletSampleApplication.kt
jakobuid: Well, we still have separate URI for pairing and request. Nothing changes in this matter. Requests have a unique uri per wallet to avoid opening app chooser every-time when Dapp sends a request.

Could you elaborate your concern?

trastvolt:
feature/deep_linking_new_flow

TalhaAli00 (approved):

PR #322, opened 2022-08-04
Feat: Add relay integration test

Elyniss:


signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -270,7 +274,7 @@ internal class SignProtocol : SignInterface, SignInterface.Websocket {
 //        wcKoinApp.close()
 //    }
 
-    private fun <T> awaitLock(codeBlock: suspend () -> T): T {
+    fun <T> awaitLock(codeBlock: suspend () -> T): T {
```
TalhaAli00, line 277: why remove the private modifier?

signSDK/sign/src/qa/kotlin/com/walletconnect/sign/di/StorageModule.kt
```diff
@@ -0,0 +1,150 @@
...
+                    databaseValue.split(",")
+                }
+
+            override fun encode(value: List<String>) = value.joinToString(separator = ",")
+        }
+    }
+
+    single<ColumnAdapter<MetaDataType, String>>(named("MetaDataType")) {
+        EnumColumnAdapter()
+    }
```
TalhaAli00, line 53: since this exact code is copied three times now, it makes more sense to move this logic into a `sharedStorageModule` file in the main directory. Then you can use Koin's `include` function with the new file and only specify the unique dependencies here

signSDK/sign/src/qa/kotlin/com/walletconnect/sign/di/StorageModule.kt
```diff
@@ -0,0 +1,150 @@
...
+        get<Database>().proposalNamespaceExtensionDaoQueries
+    }
+
+    single {
+        SequenceStorageRepository(get(), get(), get(), get(), get(), get(), get(), get(), get())
+    }
+
+    single {
+        JsonRpcHistory(get(named(DITags.RPC_STORE)), get())
+    }
```
TalhaAli00, line 149: same comment as above

TalhaAli00 (approved):

signSDK/sign/src/androidTest/kotlin/com/walletconnect/sign/utils/WCIntegrationActivityScenarioRule.kt
```diff
@@ -0,0 +1,122 @@
...
+        initialize(dappParams) {}
+        awaitLock { relay.connect { } }
+    }
+
+    override fun before() {
+        runBlocking {
+            var shouldExit = false
+            while (!shouldExit) {
+                val dappConnection = dappClient.isRelayConnectionAvailable.value
+                val walletConnection = walletClient.isRelayConnectionAvailable.value
```
jakobuid, line 80: I think it should be enough to use `dappClient.relay.eventsFlow` and `walletClient.relay.eventsFlow` to check if the `OnConnectionOpened` is received.

signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -27,6 +27,9 @@ internal class SignProtocol : SignInterface, SignInterface.Websocket {
     private val signProtocolScope =
         CoroutineScope(SupervisorJob() + Executors.newSingleThreadExecutor().asCoroutineDispatcher())
 
+    private var _isRelayConnectionAvailable: MutableStateFlow<Boolean> = MutableStateFlow(false)
+    var isRelayConnectionAvailable: StateFlow<Boolean> = _isRelayConnectionAvailable.asStateFlow()
```
jakobuid, line 31: I think we should avoid adding fields where the only reason of having them is to make test easier to write.
We should either use our current code or refactor it in the way to be more testable.
In this case I think we could use `relay.eventFlow` to check whether the connection is opened.

signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -56,6 +59,7 @@ internal class SignProtocol : SignInterface, SignInterface.Websocket {
                 wcKoinApp.modules(scarletModule(initial.relayServerUrl.addUserAgent(), jwt, initial.connectionType.toRelayConnectionType(), initial.relay))
                 signEngine = wcKoinApp.koin.get()
                 signEngine.handleInitializationErrors { error -> onError(Sign.Model.Error(error)) }
+                signEngine.isRelayConnectionAvailable.onEach { _isRelayConnectionAvailable.value = it }.launchIn(scope)
```
jakobuid, line 62: Same

signSDK/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -47,6 +47,7 @@ internal class SignEngine(
     private val _engineEvent: MutableSharedFlow<EngineEvent> = MutableSharedFlow()
     val engineEvent: SharedFlow<EngineEvent> = _engineEvent.asSharedFlow()
     private val sessionProposalRequest: MutableMap<String, WCRequestVO> = mutableMapOf()
+    val isRelayConnectionAvailable = relayer.isConnectionAvailable
```
jakobuid, line 50: Same

jakobuid (changes requested):

signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
Elyniss: I need some kind of locking mechanism to wait until sing is initialised before accessing the relay. I created new fun `blockUntilInitialized`. I had to create this due to having the manual configuration of the relay. I wonder if our consumers would need this as well.

build.gradle.kts
```diff
@@ -5,7 +5,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath("com.android.tools.build:gradle:7.2.1")
+        classpath("com.android.tools.build:gradle:7.1.3")
```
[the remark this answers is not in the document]
Elyniss: Yep, thanks for the remainder. I was able to figure it out. I needed to change the test package explicitly.

signSDK/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
TalhaAli00: Since we have the new variant, we can also create extension functions in classes that are only accessible to this test

jakobuid (approved):

signSDK/sign/src/androidTest/kotlin/com/walletconnect/sign/relay/RelayIntegrationAndroidTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+import com.walletconnect.sign.utils.WCIntegrationActivityScenarioRule
+import com.walletconnect.sign.client.Sign
+import com.walletconnect.sign.client.SignClient
+import com.walletconnect.sign.util.Logger
+import org.junit.Rule
+import org.junit.Test
+import org.junit.jupiter.api.fail
+
+class RelayIntegrationAndroidTest {
```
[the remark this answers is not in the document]
Elyniss: That's because I added env variables to describe relay and projectId

.github/workflows/ci_relay.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup Java JDK
+        uses: actions/setup-java@v3.3.0
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+          architecture: x86_64
+      - name: Instrumentation tests
```
arein, line 34: nit: why `Instrumentation`?

arein (commented):
Cool! Not so clear on what this test is actually testing. Seems like a bunch of refactor is also part of this.

Elyniss:
> Cool! Not so clear on what this test is actually testing. Seems like a bunch of refactor is also part of this.

The test itself creates a session and then sends a ping

.github/workflows/ci_relay.yml
Elyniss: In the Android context, it means that it runs on an actual device. As it is only a demo file you could rename it later on.

arein (approved):

TalhaAli00 (approved):

PR #325, opened 2022-08-08
feature/create_android_code_module

jakobuid:
Create android core module.


core/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
Elyniss, line 1: This file `core/.gitignore` is not necessary

core/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
Elyniss, line 1: This file `core/proguard-rules.pro` is not necessary

Elyniss: Also `core/consumer-rules.pro`

TalhaAli00 (commented):
Do we want to continue to call this module `Core`? Or should we change it to `Android-Core`  which will allow us to have room to create a JVM option in the future if we want?

jakobuid:
> Do we want to continue to call this module `Core`? Or should we change it to `Android-Core` which will allow us to have room to create a JVM option in the future if we want?

That was my intention. However, at some point AD neglects the `android_core` name every-time I want to use it. A module with such name cannot be find. I believe that `android` is a key words and cannot be used in the module name.

jakobuid:
> > Do we want to continue to call this module `Core`? Or should we change it to `Android-Core` which will allow us to have room to create a JVM option in the future if we want?
> 
> That was my intention. However, at some point AD neglects the `android_core` name every-time I want to use it. A module with such name cannot be find. I believe that `android` is a key words and cannot be used in the module name.

Anyway, I'd like to stress that  this module is Android specific. Do you have any other name suggestion?

foundation/src/main/java/com/walletconnect/foundation/di/ScarletModule.kt
```diff
@@ -8,7 +8,7 @@ import com.walletconnect.foundation.network.BaseRelayClient
 import com.walletconnect.foundation.network.RelayInterface
 import com.walletconnect.foundation.network.data.adapter.FlowStreamAdapter
 import com.walletconnect.foundation.network.data.service.RelayService
-import com.walletconnect.util.scope
+import com.walletconnect.foundation.util.scope
```
Elyniss, line 11: Unfortunately I'm unable to comment in appropriate line. We need to remove line 49 from this file. `.lifecycle(get())` will cause issues because we are not providing ScarletLifecycle

Elyniss (changes requested):

foundation/src/main/java/com/walletconnect/foundation/di/ScarletModule.kt
jakobuid: Nice catch I think it was already overlooked in foundation. I'm removing it.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -1,110 +1,97 @@
...
-    val navVersion = "2.4.1"
     "api"("androidx.navigation:navigation-fragment-ktx:$navVersion")
     "api"("androidx.navigation:navigation-ui-ktx:$navVersion")
 }
 
 fun DependencyHandlerScope.coroutines() {
-    val coroutinesVersion = "1.6.0"
-    "implementation"("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
+    "api"("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutinesVersion")
     "testImplementation"("org.jetbrains.kotlinx:kotlinx-coroutines-test:$coroutinesVersion")
```
Elyniss, line 23: This might be unnecessary as it already is in `coroutinesTest()`

Elyniss (changes requested):

jakobuid:
@WalletConnect/kotlin-team It was issue with my AD, thanks to Szymon I managed to change a module name to `android_core`.

TalhaAli00 (approved):

PR #332, opened 2022-08-10
feature/auth/public-api

TalhaAli00:


authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
```diff
@@ -9,12 +9,78 @@ object Auth {
...
+        )
+
+        data class AuthResponse(
+            val id: Long,
+            val message: String
+        )
+
+        sealed class Cacao: Model() {
+
+            sealed class Caip70: Cacao() {
```
Elyniss, line 56: I would remove it. I think it's obsolete as whole CACAO is CAIP70

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
```diff
@@ -9,12 +9,78 @@ object Auth {
...
+
+                data class CacaoSignature(val t: String, val s: String, val m: String?): Caip70()
```
Elyniss, line 58: If we are using layers here I would rename it to Signature only
```suggestion
                data class Signature(val t: String, val s: String, val m: String?): Caip70()
```

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
```diff
@@ -9,12 +9,78 @@ object Auth {
...
+            val aud: String,
+            val type: String?,
+            val nbf: String?,
+            val exp: String?,
+            val statement: String?,
+            val requestId: String?,
+            val resources: List<String>?,
+        ): Params()
+
+        data class Respond(val id: Long, val signature: Model.Cacao.Caip70.CacaoSignature): Params()
```
Elyniss, line 82: I think it would be better it it looked like this
```suggestion
        data class Respond(val id: Long, val signature: Model.Cacao.Signature): Params()
```

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
```diff
@@ -9,12 +9,78 @@ object Auth {
...
     sealed class Model {
+
         data class Error(val throwable: Throwable) : Model() // TODO: Should this be extracted to core for easier error handling?
 
-        sealed class Events : Model() {}
+        data class AppMetaData(
+            val name: String,
+            val description: String,
+            val url: String,
+            val icons: List<String>,
```
jakobuid, line 19: AppMedaData should include `redirect` field

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
jakobuid: I agree 👆 Cacao should be enough

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/AuthInterface.kt
```diff
@@ -1,11 +1,22 @@
 package com.walletconnect.auth.client
 
 interface AuthInterface {
-
     interface AuthDelegate {
+        fun onAuthRequest(authRequest: Auth.Model.AuthRequest)
+        fun onAuthResponse(authResponse: Auth.Model.AuthResponse)
     }
```
jakobuid, line 7: Do we want to split it to ResponderDelegate and RequesterDelegate?
I mean `onAuthRequest` will be received only by responder and `onAuthResponse` will be received only by requester.

jakobuid (changes requested):

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/AuthInterface.kt
Elyniss: I'm in favour of @jakobuid proposal

authSDK/auth/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
TalhaAli00: added

jakobuid (approved):

trastvolt:
feature/auth/public-api

Elyniss (approved):

Elyniss:
@TalhaAli00 I think we merged it into wrong branch

PR #343, opened 2022-08-12
Feature/core/copy core to android core

TalhaAli00:


jakobuid (approved):

android_core/src/main/kotlin/com/walletconnect/android_core/common/model/SymmetricKey.kt
```diff
@@ -0,0 +1,6 @@
+package com.walletconnect.android_core.common.model
+
+import com.walletconnect.foundation.common.model.Key
+
+@JvmInline
+value class SymmetricKey(override val keyAsHex: String) : Key
```
Elyniss, line 6: Why is this moved from foundation?

TalhaAli00: SymmetricKey is not used in Foundation so I moved it up into Core

PR #367, opened 2022-08-25
Feature/auth/request and respond

Elyniss:

Issue #370, opened 2022-08-26
Split Tags to specific SDKs


Elyniss:
Right now all Tags are inside android_core module which is not optimal. Instead of Enum make it an open class and have implementation inside each sdk with specific tags

PR #371, opened 2022-08-26
Auth/feature/get requests

jakobuid:
First merge: https://github.com/WalletConnect/WalletConnectKotlinV2/pull/367


android_core/src/main/kotlin/com/walletconnect/android_core/common/model/type/enums/Tags.kt
```diff
@@ -30,4 +30,8 @@ enum class Tags(val id: Int) {
 
     SESSION_PING(1114),
     SESSION_PING_RESPONSE(1115),
+
+    AUTH_REQUEST(3000),
```
TalhaAli00, line 34: I think tags should be SDK specific. We should probably convert the enum class to a sealed class and the child classes be objects

Elyniss: We have an issue for this https://github.com/WalletConnect/WalletConnectKotlinV2/issues/370

authSDK/responder/src/main/kotlin/com/walletconnect/responder/ui/accounts/AccountsFragment.kt
```diff
@@ -20,6 +20,7 @@ import com.walletconnect.responder.R
 import com.walletconnect.responder.common.ACCOUNTS_ARGUMENT_KEY
 import com.walletconnect.responder.databinding.FragmentAccountsBinding
 import com.walletconnect.responder.ui.events.ResponderEvents
+import com.walletconnect.responder.ui.request.RequestStore
```
Elyniss, line 23: Unused

signSDK/sign/src/main/kotlin/com/walletconnect/sign/storage/sequence/SequenceStorageRepository.kt
```diff
@@ -273,7 +273,7 @@ internal class SequenceStorageRepository(
     }
 
     @JvmSynthetic
-    fun deleteTempNamespacesByTopicAndRequestId(topic: String, requestId: Long) {
+    fun deleteTempNamespacesByTopicAndRequestId(requestId: Long) {
```
Elyniss, line 276: I would rename to `deleteTempNamespacesByRequestId`

TalhaAli00 (approved):

Elyniss (approved):

PR #376, opened 2022-08-30
feat: Relay Integration tests

Elyniss:


.github/workflows/ci_relay.yml
```diff
@@ -1,43 +1,43 @@
...
+
+on:
+  push:
+    branches:
+      - develop
+      - master
+  pull_request:
+    branches:
+      - develop
+      - master
```
TalhaAli00, line 11: Do we want to run this on develop?

TalhaAli00 (approved):

.github/workflows/ci_relay.yml
jakobuid: For me it depends on the time. How long does it take?

foundation/src/test/kotlin/com/walletconnect/foundation/RelayTest.kt
```diff
@@ -5,29 +5,133 @@ import com.walletconnect.foundation.di.commonModule
...
-                commonModule(),
-                cryptoModule()
-            )
+    fun `Single test`() {
+        // todo: Remove later on. Consolidate to one function below
+        `One client sends unencrypted message, second one receives it`()
+    }
+
+    @ExperimentalTime
+    // todo: Add @Test annotation later on.
```
jakobuid, line 45: Why this annotation is a problem?

jakobuid (approved):

.github/workflows/ci_relay.yml
Elyniss: @jakobuid Most recent took 1,5 min: https://github.com/WalletConnect/WalletConnectKotlinV2/actions/runs/2962097261

foundation/src/test/kotlin/com/walletconnect/foundation/RelayTest.kt
Elyniss: I moved this test to a separate function so I can showcase current issues on our rs-relay, which are not found in ts-relay. Later on we should add `@Test` annotation here and remove tests: `N consequent tests` and `Single test`

.github/workflows/ci_relay.yml
Elyniss: And another sample which took 2,5 min: https://github.com/WalletConnect/WalletConnectKotlinV2/actions/runs/2962142651

Elyniss: In conclusion between 1,5 min to 3 min. Also, I added caching to our unit tests, which shorten the execution time from ~6-7 min to 2~3 min.

PR #406, opened 2022-09-13
Fix/relay tests

jakobuid:
- Fix for RelayTests
- Refactor handling result from Relay methods


Elyniss (approved):

foundation/src/main/kotlin/com/walletconnect/foundation/network/BaseRelayClient.kt
```diff
@@ -22,7 +22,7 @@ import org.koin.core.KoinApplication
 abstract class BaseRelayClient : RelayInterface {
     private var foundationKoinApp: KoinApplication = KoinApplication.init()
     lateinit var relayService: RelayService
-    private var logger: Logger
+    protected open lateinit var logger: Logger
```
TalhaAli00, line 25: I don't think you need open. protected should be enough

TalhaAli00 (approved):

PR #407, opened 2022-09-13
Fix for session request parsing

jakobuid:


androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/BaseJsonRpcInteractor.kt
```diff
@@ -212,8 +212,11 @@ open class BaseJsonRpcInteractor(
 
     private suspend fun handleRequest(clientJsonRpc: ClientJsonRpc, topic: Topic, decryptedMessage: String) {
         if (jsonRpcHistory.setRequest(clientJsonRpc.id, topic, clientJsonRpc.method, decryptedMessage)) {
+
             serializer.deserialize(clientJsonRpc.method, decryptedMessage)?.let { params ->
+
                 _clientSyncJsonRpc.emit(WCRequest(topic, clientJsonRpc.id, clientJsonRpc.method, params))
+
             } ?: handleError("JsonRpcInteractor: Unknown request params")
```
TalhaAli00, line 220: seems like excessive lines, I think it was fine before

TalhaAli00 (approved):

PR #416, opened 2022-09-20
Add logs for client id

jakobuid:


arein (approved):
Thank you!

foundation/src/main/kotlin/com/walletconnect/foundation/crypto/data/repository/BaseJwtRepository.kt
```diff
@@ -34,12 +33,13 @@ abstract class BaseJwtRepository: JwtRepository {
         return Base64.getUrlEncoder().withoutPadding().encodeToString(signature)
     }
 
-    override fun generateJWT(serverUrl: String): String {
+    override fun generateJWT(serverUrl: String, getIssuer: (String) -> Unit): String {
```
TalhaAli00, line 36: do we have to have this in the common impl? Can't this logic just stay in the base class definition?

TalhaAli00, line 36: let's default the lamda to an empty function

TalhaAli00 (approved):

foundation/src/main/kotlin/com/walletconnect/foundation/crypto/data/repository/BaseJwtRepository.kt
jakobuid: It's already defaulted in the interface

Elyniss (approved):

PR #465, opened 2022-11-07
Add a new extra field for an artifact description

jakobuid:


Elyniss (approved):

buildSrc/src/main/kotlin/publish-module-android.gradle.kts
```diff
@@ -38,7 +38,7 @@ afterEvaluate {
 
                 pom {
                     name.set("WalletConnect ${requireNotNull(extra.get(KEY_SDK_NAME))}")
-                    description.set("Sign SDK for WalletConnect.")
+                    description.set("${requireNotNull(extra.get(KEY_SDK_DESCRIPTION))}")
```
TalhaAli00, line 41: instead of a descriptor, could we not do something like the example below and use the SDK name property?

description.set("${requireNotNull(extra.get(KEY_SDK_NAME)) for WalletConnect}")

jakobuid: Make sense, I like it

TalhaAli00 (approved):

Issue #424, opened 2022-10-05
Add Proguard rules to SDKs


TalhaAli00:
add rules to

- [ ] Sign SDK
- [ ] Auth SDK
- [ ] Android Core SDK

PR #480, opened 2022-11-10
Feature/core/add proguard

TalhaAli00:
#424


androidCore/impl/proguard-rules.pro
```diff
@@ -0,0 +1,6 @@
+#-keep class com.walletconnect.android.impl.** { *; }
+#-keep class com.walletconnect.utils.** { *; }
```
jakobuid, line 2: q: Shouldn't we use here `keepnames` to allow shrinking but no obfuscation for classes and their members? What do you think?

sign/sdk/proguard-rules.pro
```diff
@@ -0,0 +1,5 @@
+-keep public class com.walletconnect.sign.** { *; }
+
+-repackageclasses 'com.walletconnect.sign'
+-allowaccessmodification
+-keeppackagenames doNotKeepAThing
```
jakobuid, line 5: Why do we need those in here but not in the chad and auth sdks?

TalhaAli00: I only had to add this because there was a conflict with the packaging. I think we should only use stuff as needed

androidCore/impl/proguard-rules.pro
TalhaAli00: When I did keepnames, the dependency was failing. So I changed it to keep

jakobuid (approved):

Elyniss (approved):

PR #485, opened 2022-11-15
Improve error handling in keychain

jakobuid:


androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/storage/KeyChain.kt
```diff
@@ -24,9 +24,8 @@ internal class KeyChain(private val sharedPreferences: SharedPreferences) : KeyS
         sharedPreferences.edit().putString(tag, keys).apply()
     }
 
-    @Throws(InternalError::class)
-    override fun getKeys(tag: String): Pair<String, String> {
-        val concatKeys = sharedPreferences.getString(tag, null) ?: throw InternalError("Unable to find keys")
+    override fun getKeys(tag: String): Pair<String, String>? {
+        val concatKeys = sharedPreferences.getString(tag, null) ?: return null
```
TalhaAli00, line 28: we need an additional check to see if `concatKeys` is null and do a early return

androidCore/sdk/src/main/kotlin/com/walletconnect/android/pairing/client/PairingInterface.kt
```diff
@@ -18,14 +17,14 @@ interface PairingInterface {
...
 
     //  idea: --- I think those below shouldn't be accessible by SDK consumers.
-    fun activate(topic: String, onError: (Core.Model.Error) -> Unit = {})
+    fun activate(activate: Core.Params.Activate, onError: (Core.Model.Error) -> Unit = {})
 
-    fun updateExpiry(topic: String, expiry: Expiry, onError: (Core.Model.Error) -> Unit = {})
+    fun updateExpiry(updateExpiry: Core.Params.UpdateExpiry, onError: (Core.Model.Error) -> Unit = {})
 
-    fun updateMetadata(topic: String, metadata: Core.Model.AppMetaData, metaDataType: AppMetaDataType, onError: (Core.Model.Error) -> Unit = {})
+    fun updateMetadata(updateMetadata: Core.Params.UpdateMetadata, onError: (Core.Model.Error) -> Unit = {})
```
TalhaAli00, line 27: Let's keep the functions with the string parameters and add deprecated to the old functions. That way we don't break functionality for current implementors

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/storage/KeyChain.kt
jakobuid: I don't get what you mean. I think this: `sharedPreferences.getString(tag, null) ?: return null` prevents us from having a null value

androidCore/sdk/src/main/kotlin/com/walletconnect/android/pairing/client/PairingInterface.kt
jakobuid: Good point. Although, It makes sense only for the disconnect method since `activate`, `updateExpiry`, `updateMetadata`, and `register` will be hidden soon within the new internal interface.

androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -74,7 +74,10 @@ internal class JsonRpcInteractor(
 
         val requestJson = serializer.serialize(payload) ?: return onFailure(IllegalStateException("JsonRpcInteractor: Unknown result params"))
 
+        Logger.error("kobe; Request: $requestJson")
```
Elyniss, line 77: lol

androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -103,6 +106,9 @@ internal class JsonRpcInteractor(
 
         val jsonResponseDO = response.toJsonRpcResponse()
         val responseJson = serializer.serialize(jsonResponseDO) ?: return onFailure(IllegalStateException("JsonRpcInteractor: Unknown result params"))
+
+        Logger.error("kobe; Response: $response")
```
Elyniss, line 110: same

androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -219,6 +225,8 @@ internal class JsonRpcInteractor(
                     val topic = Topic(relayRequest.subscriptionTopic)
                     val message = chaChaPolyCodec.decrypt(topic, relayRequest.message)
 
+                    Logger.error("kobe; Message: $message")
```
Elyniss, line 228: same

androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -139,7 +145,7 @@ internal class JsonRpcInteractor(
 
         try {
             publishJsonRpcResponse(request.topic, irnParams, result, envelopeType = envelopeType, participants = participants,
-                onFailure = { error -> Logger.error("Cannot send the response, error: $error") })
+                onFailure = { error -> handleError("Cannot send the response, error: $error") })
```
Elyniss, line 148: should we add it as well to `respondWithParams` ?

jakobuid: No, we don't need it there. Although, we do need a callback to let developers know that sending the response failed.

Elyniss (approved):

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/storage/KeyChain.kt
TalhaAli00: Ya, you're right. I was thinking about the case where the key does exist but the result comes back as null. But the return should catch that case. There's still an edge case where the key exists and is not null but we can't split the key correctly. But we it unlikely for that to happen

androidCore/sdk/src/main/kotlin/com/walletconnect/android/pairing/client/PairingInterface.kt
TalhaAli00: that's true for when we remove them. But at the moment, we don't want to break these public functions right now. We can deprecate `activate`, `updateExpiry`, `updateMetadata`, and `register` and make them noop functions and then phase them out instead of breaking people's builds

TalhaAli00 (approved):

PR #505, opened 2022-11-25
Add fix to avoid breaking data flow when exception occurs

jakobuid:


androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -143,9 +141,9 @@ internal class JsonRpcInteractor(
 
         try {
             publishJsonRpcResponse(request.topic, irnParams, result, envelopeType = envelopeType, participants = participants,
-                onFailure = { error -> handleError("Cannot send the response, error: $error") })
+                onFailure = { error -> handleError("Cannot send the responseWithSuccess, error: ${error.stackTraceToString()}") })
         } catch (e: Exception) {
-            handleError(e.message ?: String.Empty)
+            handleError("responseWithSuccess; ${e.stackTraceToString()}")
```
TalhaAli00, line 146: I don't think we can say that the publish was sent here so the `respondWithSucess` might be misleading. Maybe we change it to `publishFailure` or something

androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -163,11 +161,11 @@ internal class JsonRpcInteractor(
         try {
             publishJsonRpcResponse(request.topic, irnParams, jsonRpcError, envelopeType = envelopeType, participants = participants,
                 onFailure = { failure ->
-                    handleError("Cannot respond with error: $failure")
+                    handleError("Cannot send respondWithError: ${failure.stackTraceToString()}")
                     onFailure(failure)
                 })
         } catch (e: Exception) {
-            handleError(e.message ?: String.Empty)
+            handleError("respondWithError; ${e.stackTraceToString()}")
```
TalhaAli00, line 168: similar message as above

androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/json_rpc/domain/JsonRpcInteractor.kt
jakobuid: Makes sense to me

Elyniss (approved):

TalhaAli00 (approved):

PR #506, opened 2022-11-28
fix: allow logs in debug mode

jakobuid:


androidCore/sdk/src/main/kotlin/com/walletconnect/android/relay/RelayClient.kt
```diff
@@ -56,7 +59,7 @@ object RelayClient : BaseRelayClient(), RelayConnectionInterface {
         get() =
             eventsFlow
                 .onEach { event: Relay.Model.Event ->
-                    logger.log("$event")
+                    logger.log("timber: $event")
```
Elyniss, line 62: I would prefer more domain scoped tag instead of `timber:`. How about `relayEvent:` or none ?

jakobuid: Oh, I left it by accident, there should not be any tag

androidCore/sdk/src/main/kotlin/com/walletconnect/android/relay/RelayClient.kt
```diff
@@ -39,6 +40,8 @@ object RelayClient : BaseRelayClient(), RelayConnectionInterface {
         }
 
         logger = wcKoinApp.koin.get(named(AndroidCommonDITags.LOGGER))
+        plantTimber()
```
Elyniss, line 43: How about we move and call `plantTimber` inside `com.walletconnect.android.internal.common.di.CommonModule.kt` ?

I feel like it's in weird place right now

jakobuid: I don't mind moving it there

Elyniss (approved):

TalhaAli00 (approved):

PR #510, opened 2022-11-30
Feature/core/create bom

TalhaAli00:


androidCore/bom/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest package="com.walletconnect.android.bom" />
```
Elyniss, line 2: Might not be true but we can probably get rid of this file if we add `namespace` property to build.gradle

```
android {
	namespace = "com.walletconnect.android.bom"
}
```

sign/wallet/build.gradle.kts
```diff
@@ -44,10 +44,13 @@ dependencies {
     implementation(project(":sign:samples_common"))
 
     debugImplementation(project(":sign:sdk"))
-    releaseImplementation("com.walletconnect:sign:2.2.0")
+//    releaseImplementation("com.walletconnect:sign:2.2.0")
 
     debugImplementation(project(":androidCore:sdk"))
-    releaseImplementation("com.walletconnect:android-core:1.4.0")
```
Elyniss, line 50: should we add bom to other project `releaseImplementation`s as well?

androidCore/sdk/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.walletconnect.android.api">
```
Elyniss, line 3: nice catch!

androidCore/bom/src/main/AndroidManifest.xml
TalhaAli00: removed the manifest

sign/wallet/build.gradle.kts
TalhaAli00: updated all of the samples to use bom. BOM will need to be built locally until we release

jakobuid (approved):

Issue #517, opened 2022-12-05
Remove useless dependency


stosabon:
The problem is with `restEasyJava` dependency. The only usage of this library is `javax.ws.rs.core.UriBuilder`. But under the hood `restEasyJava` uses org.jboss.spec.javax.annotation instead of default javax.annotation and it makes your library incompatible with all projects using default `javax.annotation` So I want to change using of `restEasyJava` to direct `javax.ws.rs` library. Don't see a way how to contribute these changes. I've already checked that it works, all tests are passed.
Don't see any reason to use `org.jboss.resteasy`

jakobuid:
Hey @stosabon can you create a PR with your proposition?

stosabon:
@jakobuid remote: Permission to WalletConnect/WalletConnectKotlinV2.git denied to stosabon. unable to access 'https://github.com/WalletConnect/WalletConnectKotlinV2.git/': The requested URL returned error: 403

stosabon:
@jakobuid

PR #518, opened 2022-12-05
Remove unused dependencies

stosabon:
https://github.com/WalletConnect/WalletConnectKotlinV2/issues/517


stosabon:
@jakobuid I think the problem of failing build is with getting `testProjectId` from secrets in `RelayTest` .  I see that it is empty in test. It pass with real projectID.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -105,8 +105,9 @@ fun DependencyHandlerScope.multibaseJava() {
     "api"("com.github.multiformats:java-multibase:$multibaseVersion") //https://mvnrepository.com/artifact/com.github.multiformats/java-multibase/1.1.0 vulnerability detected with library
 }
 
-fun DependencyHandlerScope.restEasyJava() {
-    "implementation"("org.jboss.resteasy:resteasy-jaxrs:$restEasyVersion")
+fun DependencyHandlerScope.wsRestJava() {
+    "implementation"("javax.ws.rs:javax.ws.rs-api:$wsRestJavaVersion")
+    "testImplementation"("org.glassfish.jersey.core:jersey-common:2.24")
```
TalhaAli00, line 110: It looks like there's a vulnerability in the jersey dependency for this version (https://mvnrepository.com/artifact/org.glassfish.jersey.core/jersey-common/2.24). Remove the jersey dependency

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -105,8 +105,9 @@ fun DependencyHandlerScope.multibaseJava() {
     "api"("com.github.multiformats:java-multibase:$multibaseVersion") //https://mvnrepository.com/artifact/com.github.multiformats/java-multibase/1.1.0 vulnerability detected with library
 }
 
-fun DependencyHandlerScope.restEasyJava() {
-    "implementation"("org.jboss.resteasy:resteasy-jaxrs:$restEasyVersion")
+fun DependencyHandlerScope.wsRestJava() {
+    "implementation"("javax.ws.rs:javax.ws.rs-api:$wsRestJavaVersion")
```
TalhaAli00, line 109: This dependency also has a potential vulnerability. It has also been moved. Replace this dependency with `implementation("jakarta.ws.rs:jakarta.ws.rs-api:3.1.0")`

buildSrc/src/main/kotlin/Dependencies.kt
stosabon: jersey is needed for tests, but yeah, I need to upgrade version also
Otherwise, it will fail test with "java.lang.ClassNotFoundException: Provider for jakarta.ws.rs.ext.RuntimeDelegate cannot be found"

TalhaAli00 (approved):

jakobuid (approved):

PR #526, opened 2022-12-09
feat: auth multi account

jakobuid:


androidCore/impl/src/main/kotlin/com/walletconnect/android/impl/utils/Validator.kt
```diff
@@ -0,0 +1,26 @@
+package com.walletconnect.android.impl.utils
+
+object Validator {
```
TalhaAli00, line 3: Can we name this `CoreValidator`?

TalhaAli00 (approved):

PR #549, opened 2022-12-23
Feature/push/add echo client

TalhaAli00:


androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/model/EchoResponse.kt
```diff
@@ -0,0 +1,32 @@
...
+        @Json(name = "name")
+        val name: String
+    )
+
+    @JsonClass(generateAdapter = true)
+    data class Field(
+        @Json(name = "description")
+        val description: String,
+        @Json(name = "field")
+        val `field`: String,
```
jakobuid, line 28: Just curious: why is there single quotation mark?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/network/EchoService.kt
```diff
@@ -0,0 +1,18 @@
...
+import retrofit2.http.DELETE
+import retrofit2.http.POST
+import retrofit2.http.Path
+
+interface EchoService {
+
+    @POST("893ad3af-9515-42d6-aaa2-06b53494b9a5/clients")
+    suspend fun register(@Body echoClientsBody: EchoBody): Response<EchoResponse>
+
+    @DELETE("893ad3af-9515-42d6-aaa2-06b53494b9a5/clients/{clientId}")
```
jakobuid, line 16: This "893ad3af-9515-42d6-aaa2-06b53494b9a5" looks like some static value. What is that?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import retrofit2.Retrofit
+
+internal object EchoClient: EchoInterface {
+    private val echoService by lazy { wcKoinApp.koin.get<Retrofit>().create(EchoService::class.java) }
+    private val clientId by lazy { requireNotNull(wcKoinApp.koin.get<SharedPreferences>().getString(EchoInterface.KEY_CLIENT_ID, null)) }
+    private const val SUCCESS_STATUS = "SUCCESS"
+
+    internal fun initialize() {}
```
jakobuid, line 21: What's the purpose of this method? When EchoClient should be initialized? Should it be a part of the CoreClient?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.walletconnect.android.internal.common.scope
+import com.walletconnect.android.internal.common.wcKoinApp
+import com.walletconnect.foundation.common.model.Topic
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import retrofit2.Retrofit
+
+internal object EchoClient: EchoInterface {
+    private val echoService by lazy { wcKoinApp.koin.get<Retrofit>().create(EchoService::class.java) }
+    private val clientId by lazy { requireNotNull(wcKoinApp.koin.get<SharedPreferences>().getString(EchoInterface.KEY_CLIENT_ID, null)) }
```
jakobuid, line 18: Can't we create a echo koin module that'll be initialized inside the initialize function of EchoClient?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
```diff
@@ -0,0 +1,75 @@
...
+                } else {
+                    onError(IllegalArgumentException(response.errorBody()?.string()))
+                }
+            }
+        } catch (e: Exception) {
+            onError(e)
+        }
+    }
+
+    override fun unregister(onSuccess: () -> Unit, onError: (Throwable) -> Unit) {
```
jakobuid, line 45: When unregister should be called?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
```diff
@@ -0,0 +1,75 @@
...
+                } else {
+                    onError(IllegalArgumentException(response.errorBody()?.string()))
+                }
+            }
+        } catch (e: Exception) {
+            onError(e)
+        }
+    }
+
+    override fun decryptMessage(topic: String, message: String, onSuccess: (String) -> Unit, onError: (Throwable) -> Unit) {
```
jakobuid, line 65: Why decryptMessage method is needed here?  What's the purpose of it? There is already one in the PushWalletInterface.

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushWalletInterface.kt
```diff
@@ -25,5 +25,7 @@ interface PushWalletInterface {
 
     fun delete(params: Push.Wallet.Params.Delete, onError: (Push.Model.Error) -> Unit)
 
-    fun decryptMessage(params: Push.Wallet.Params.DecryptMessage, onSuccess: (Push.Model.Message) -> Unit, onError: (Push.Model.Error) -> Unit)
+    fun registerFcmToken(params: Push.Wallet.Params.FcmToken, onSuccess: () -> Unit, onError: (Push.Model.Error) -> Unit)
```
jakobuid, line 28: Isn't it enough to just call a register method from EchoClient? Instead of just wrapping it in the PushWalletProtocol.

Also, from what I read in specs, register is not a part of the PushWalletClient.

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushWalletProtocol.kt
```diff
@@ -36,8 +34,6 @@ class PushWalletProtocol : PushWalletInterface {
                 walletEngineModule(),
```
jakobuid, line 34: // TODO: Commented out until we merge PR to handle multiple versions of dependnecy
//                pushCommonModule(),
//                cryptoModule(),

Let's clean this up since the PR responsible for consolidating androidCore-impl into androidCore-sdk is merged

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushWalletProtocol.kt
```diff
@@ -95,12 +91,20 @@ class PushWalletProtocol : PushWalletInterface {
...
+
+        echoClient.register(params.fcmAccessToken, onSuccess) { error ->
+            onError(Push.Model.Error(error))
+        }
+    }
+
+    override fun decryptMessage(params: Push.Wallet.Params.DecryptMessage, onSuccess: (String) -> Unit, onError: (Push.Model.Error) -> Unit) {
+        val echoClient = wcKoinApp.koin.get<EchoInterface>()
+
+        echoClient.decryptMessage(params.topic, params.encryptedMessage, onSuccess) { error ->
```
jakobuid, line 105: I don't see a need for decrypt method inside EchoClient

sign/wallet/build.gradle.kts
```diff
@@ -44,13 +45,16 @@ dependencies {
     implementation(project(":sign:samples_common"))
 
     debugImplementation(project(":androidCore:sdk"))
-    debugImplementation(project(":sign:sdk"))
+    debugImplementation(project(":web3:wallet"))
```
jakobuid, line 48: Why this change?

Also, I think it would be great to have Push integration in our new Web3Wallet app, so that people could actually test the whole flow.

sign/wallet/src/main/kotlin/com/walletconnect/wallet/WalletFirebaseMessagingService.kt
```diff
@@ -0,0 +1,68 @@
...
+import androidx.core.app.NotificationCompat
+import com.google.firebase.messaging.RemoteMessage
+import com.walletconnect.push.common.Push
+import com.walletconnect.push.wallet.client.PushMessageService
+import com.walletconnect.wallet.ui.host.WalletSampleActivity
+import kotlin.random.Random
+import kotlin.random.nextUInt
+
+@SuppressLint("MissingFirebaseInstanceTokenRefresh")
+class WalletFirebaseMessagingService: PushMessageService() {
```
jakobuid, line 21: This idea of creating our own messaging service is very nice. I like it, good job.

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -86,4 +88,11 @@ fun coreAndroidNetworkModule(serverUrl: String, jwt: String, connectionType: Con
         ConnectivityState(androidApplication())
     }
 
+    single {
```
Elyniss, line 91: As there will be another Retrofit instance in Chat SDk for Keyserver. Could we use named dependency here?

build.gradle.kts
```diff
@@ -16,6 +16,7 @@ buildscript {
         classpath("org.jetbrains.dokka:dokka-core:$dokkaVersion")      // TODO: Leave version until AGP 7.3 https://github.com/Kotlin/dokka/issues/2472#issuecomment-1143604232
         classpath("org.jetbrains.dokka:dokka-gradle-plugin:$dokkaVersion")
         classpath("com.squareup.sqldelight:gradle-plugin:$sqlDelightVersion")
+        classpath("com.google.gms:google-services:4.3.14")
```
Elyniss, line 19: Let's extract version

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/model/EchoResponse.kt
TalhaAli00: since field is a Kotlin keyword, the single quotation marks are needed to have a variable of the same name

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/network/EchoService.kt
TalhaAli00: this is the FCM endpoint for echo server

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
TalhaAli00: I kept initialize but it makes sense to remove it now

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
TalhaAli00: unregister is for wallets to tell Echo server to not send anymore notifications

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
TalhaAli00: That's a good callout. I've removed it from EchoClient

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushWalletProtocol.kt
TalhaAli00: removed

sign/wallet/build.gradle.kts
TalhaAli00: missed this during rebase. Fixed

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
TalhaAli00: done

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushWalletProtocol.kt
TalhaAli00: moved logic to PushWallet client

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
TalhaAli00: What do you mean by having an echo koin module? A koin module defines dependencies

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushWalletInterface.kt
TalhaAli00: You're right about not being in the specs. I've removed this function and reverted the service to use `CoreClient.Echo`

Elyniss (approved):

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
jakobuid: It's not clear to me how developers initialize EchoClient. If it's by calling the init function we could get rid of lazy koin initialization and create a koin module for that. Such a module could be called inside the init function.

I want to avoid any potential race conditions on initialization and assure the ride order on components initialization.

jakobuid: Ok, I've just noticed that the initialization of EchoClient happens in CoreClient. This makes sense, so could we add EchoService and clientId to Koin already inside CoreClient?
Does it make sense to you?

sign/wallet/src/main/kotlin/com/walletconnect/wallet/WalletFirebaseMessagingService.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    override fun registeringFailed(token: String, throwable: Throwable) {
+        Log.d(TAG, token)
+    }
+
+    override fun onDecryptedMessage(decryptedMessage: String, originalMessage: RemoteMessage) {
+        Log.d(TAG, decryptedMessage)
+    }
+
+    override fun onUnencryptedMessage(message: Push.Model.Message, originalMessage: RemoteMessage) {
```
jakobuid, line 40: I think from the developer's perspective it shouldn't matter whether the message was decrypted by the SDK or was unencrypted when received.
It'd be more convenient to have one callback such as onMessageReceived that has only a Push.Model.Message as a parameter.
The output of decrypting the encrypted message should be the same as the unencrypted message received by the SDK.

androidCore/sdk/src/main/kotlin/com/walletconnect/android/echo/EchoClient.kt
TalhaAli00: Makes sense. I've added a Echo module

sign/wallet/src/main/kotlin/com/walletconnect/wallet/WalletFirebaseMessagingService.kt
TalhaAli00: That's a valid point. I've consolidated the two functions and handle both cases now

push/sdk/src/main/kotlin/com/walletconnect/push/wallet/client/PushMessageService.kt
```diff
@@ -0,0 +1,63 @@
...
+                onError(e, message)
+            }
+        }
+    }
+
+    abstract fun newToken(token: String)
+
+    abstract fun registeringFailed(token: String, throwable: Throwable)
+
+    abstract fun ondMessage(message: Push.Model.Message, originalMessage: RemoteMessage)
```
jakobuid, line 56:
```suggestion
    abstract fun onMessage(message: Push.Model.Message, originalMessage: RemoteMessage)
```

jakobuid (approved):
I left one last comment about typo, approved 👍

PR #567, opened 2023-01-05
Add onSuccess callback to auth and web3wallet

jakobuid:


build.gradle.kts
```diff
@@ -16,6 +16,8 @@ buildscript {
         classpath("org.jetbrains.dokka:dokka-core:$dokkaVersion")      // TODO: Leave version until AGP 7.3 https://github.com/Kotlin/dokka/issues/2472#issuecomment-1143604232
         classpath("org.jetbrains.dokka:dokka-gradle-plugin:$dokkaVersion")
         classpath("com.squareup.sqldelight:gradle-plugin:$sqlDelightVersion")
+        classpath("com.google.gms:google-services:4.3.14")
+        classpath("com.google.firebase:firebase-crashlytics-gradle:2.9.2")
```
TalhaAli00, line 20: good idea to add Crashlytics to the wallet

samples/web3wallet/src/main/kotlin/com/walletconnect/web3/wallet/Web3WalletApplication.kt
```diff
@@ -23,7 +24,12 @@ class Web3WalletApplication : Application() {
...
         )
 
-        CoreClient.initialize(relayServerUrl = serverUrl, connectionType = ConnectionType.AUTOMATIC, application = this, metaData = appMetaData) { error ->
+        CoreClient.initialize(
+            relayServerUrl = serverUrl,
+            connectionType = ConnectionType.AUTOMATIC,
+            application = this,
+            metaData = appMetaData
+        ) { error ->
             Log.e("CoreClient", error.throwable.stackTraceToString())
```
TalhaAli00, line 33: instead of doing logs like this, we should use Crashlytics and report the exception 
https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=android#log-excepts

TalhaAli00: We should add the `onError` callback and log the exception to Crashlytics to all calls

jakobuid: Good catch, I'm going to replace all the occurrences

Elyniss (approved):

TalhaAli00 (approved):

PR #568, opened 2023-01-05
The PairingParams classes were passed to the JsonRpcSerializer instea…

TalhaAli00:
…d of PairingRpc classes. Fixed by having the correct classes for serializing and deserializing. Also simplified check when checking payload in JsonRpcSerializer


androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/json_rpc/data/JsonRpcSerializer.kt
```diff
@@ -33,7 +34,7 @@ class JsonRpcSerializer(
             payload is JsonRpcResponse.JsonRpcError -> trySerialize(payload)
             serializerEntries.any { type: KClass<*> ->
                 payloadType = type
-                payload::class == type
+                type.safeCast(payload) != null
```
Elyniss, line 37: what's the difference here? Were there any issues before?

TalhaAli00: Not really an issue. The safeCast actually does the casting instead of looking at just the type. 

This would also support casting to parents without breaking vs an exact type matching so i thought it would be safer to do it this way.

Elyniss (approved):

jakobuid (approved):
It's great to have one less bug on our backlog 🤝

jemshit:
I still get this error from TrustWallet, do they need to upgrade their dep. version ?

jemshit:
With having this error, i can see `CoreClient.Pairing.getPairings()` returning empty list, but `SignClient.getListOfActiveSessions()` not empty. Is this expected or is this because of this bug? Shouldn't deleting Pair trigger delete all sessions for that pair?

jemshit:
> I still get this error from TrustWallet, do they need to upgrade their dep. version ?

this was because of my dep. version. Still deleting Pair doesn't delete Sessions o.O

PR #571, opened 2023-01-09
✨ Make extending expiration date customizable

mitsinsar:


jakobuid (commented):
Hey @mitsinsar thanks for another PR!

Your suggestion looks interesting, before reviewing could you tell us more about your use case in which you need to extend session with different expiry than 7days?

We'll definitely consider it and discuss internally with other teams to make this feature available across all platforms.

jakobuid:
> Hello @jakobuid! You are welcome 🙏
> 
> I -as a user- keep my single session active all the time. With WC V2, session lifetime will be `7 days` as default and `extend` function will extend it only `1 day (DAY_IN_SECONDS)`. I thought it would be better to extend it more than one day, so that I can keep my session more than a week. Instead of setting TTL to `WEEK_IN_SECONDS`, I wanted to make it customizable. It would be nice to have a feature to sync pair & session expiration dates.

The current implementation of the extend method will increase a session's expiry by 7 days every time you call it. If you look at this line in SignEngine (https://github.com/WalletConnect/WalletConnectKotlinV2/blob/develop/sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt#L424) you can see how a new expiration is calculated.

Then, it is sent to the other peer with the JsonRpc publish request (https://github.com/WalletConnect/WalletConnectKotlinV2/blob/develop/sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt#L426), so that the other peer is informed about the new expiry value.

The TTL though is for a server to know for how long the published message has to be persisted.

Is it something that satisfies your use case?

mitsinsar:
Oh, I completely misunderstood then. Thank you so much for clarification and sorry for taking your time. 🙏

PR #580, opened 2023-01-12
fix: class duplication error

jakobuid:
- Fix for a class duplication error
- Update minor dependencies
- Remove duplicated dependencies in gradle files


buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -128,7 +128,7 @@ fun DependencyHandlerScope.androidXTest() {
     "testImplementation"("androidx.test:core-ktx:$androidxTestVersion")
 
     "androidTestImplementation"("androidx.test:core-ktx:$androidxTestVersion")
-    "androidTestImplementation"("androidx.test:runner:$androidxTestVersion")
+    "androidTestImplementation"("androidx.test:runner:1.5.2")
```
Elyniss, line 131: Why is here a specific value?

jakobuid: Because the newest version for runners differs from the core one

Elyniss (approved):

androidCore/sample/build.gradle.kts
```diff
@@ -19,6 +19,8 @@ android {
     buildTypes {
         release {
```
TalhaAli00, line 20: Maybe we don't need this sample. I think we can get rid of it

sign/sdk/build.gradle.kts
```diff
@@ -65,7 +65,6 @@ dependencies {
     debugImplementation(project(":androidCore:sdk"))
     releaseImplementation("com.walletconnect:android-core:$CORE_VERSION")
 
-    moshiKsp()
```
TalhaAli00, line 68: Why was this removed?

TalhaAli00 (approved):

androidCore/sample/build.gradle.kts
jakobuid: Yea that's true we could

sign/sdk/build.gradle.kts
jakobuid: Because while testing sdks with updated dependencies it seemed to be unnecessary here

Issue #556, opened 2023-01-02
Connection state change is always reported as false on some devices


kamilargent:
**Describe the bug**
On some devices, the connection state is constantly reported as false - even though the user has an internet connection.

**SDK Version**
 - Client: Android BOM
 - Version 1.0.1

**To Reproduce**
1. Setup a device with at least 2 active networks e.g. Bluetooth and LTE
2. Try to initialize the SDK

**Expected behavior**
A user is able to interact with SDK when in reality he can't do anything

**Device (please complete the following information):**
 - Device: Samsung Galaxy S10
 - OS: Android 12
 - Browser: Chrome

**Additional context**
As a result in the logs we can constantly see `onConnectionStateChange(ConnectionState(isAvailable=false))`

The logic in `ConnectivityState` to check the network connectivity is not covering all cases.
It is common for the user to have multiple active networks, e.g. when plugging in Bluetooth headphones on Samsung devices. Sometimes it can even happen that there are multiple networks with internet capabilities. Because of that iterating over all networks is necessary to check internet capabilities.

The `isConnected` could be modified to sth like this:
```
    private fun isConnected(): Boolean = connectivityManager.run {
        allNetworks.any { network ->
            getNetworkCapabilities(network)?.run {
                hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) && hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
            } ?: false
        }
    }
```
This is just an example: the `allNetworks` is deprecated since API 31 since it's not efficient - it can be achieved in multiple ways using data reported by network callback.

PS. The `NetworkRequest` can be improved by listening to only networks that have internet capabilities - it should have better performance.

jakobuid:
Hey, @kamilargent thanks for reporting.

I did some digging and I was indeed at some point able to reproduce your issue on the Samsung device, however, my conclusions are slightly different though.

On Samsung devices whenever you enable the Bluetooth connection, the Samsung's Bluetooth features dialog pops up, which causes the current application to be backgrounded, which is the reason of receiving the `onConnectionStateChange(ConnectionState(isAvailable=false))` because SDK closes the wss connection in AUTOMATIC when that happens.

However, every time when I closed the Samsung's Bluetooth dialog the wss connection where opened again.

For more context, you can have a look on the video included below.

It'd be great if you could attached a video from your side where I can see the app and logs from logcat at the same time.

[...]

PR #581, opened 2023-01-13
fix:connectivity state detection

jakobuid:
As discussed in https://github.com/WalletConnect/WalletConnectKotlinV2/issues/556 the connectivity state detects the Bluetooth connection as an active network on Samsung devices.

The issue above couldn't be fully reproduced and the following changes seem to resolve that problem.

@kamilargent please participate in the review process as an author of the reported bug. Once we agree on the solution our team will release a SNAPSHOT artifact to continue with manual testing.

What's changed:
- Whenever a network request callback has triggered a network passed by a particular method is checked for its transport and capabilities
- The network request build defines the desired network capabilities and transport

### DIFF
--- a/androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/connection/ConnectivityState.kt
+++ b/androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/connection/ConnectivityState.kt
@@ -18,38 +18,46 @@ internal class ConnectivityState(context: Context) {
     val isAvailable: StateFlow<Boolean> = _isAvailable.asStateFlow()
 
     private val callback = object : ConnectivityManager.NetworkCallback() {
+
         override fun onAvailable(network: Network) {
-            updateIfStateChanged()
+            updateIfStateChanged(network)
         }
 
         override fun onCapabilitiesChanged(network: Network, networkCapabilities: NetworkCapabilities) {
-            updateIfStateChanged()
+            updateIfStateChanged(network)
         }
 
         override fun onLost(network: Network) {
-            updateIfStateChanged()
+            updateIfStateChanged(network)
         }
     }
 
-    private fun updateIfStateChanged() {
-        val isNowConnected = isConnected()
-        val wasConnected = _isAvailable.value
+    private fun updateIfStateChanged(network: Network) {
+        val isNowConnected = isNowConnected(network)
 
-        if (!wasConnected && isNowConnected) {
+        if (isNowConnected) {
             _isAvailable.compareAndSet(expect = false, update = true)
-        } else if (wasConnected && !isNowConnected) {
+        } else {
             _isAvailable.compareAndSet(expect = true, update = false)
         }
     }
 
-    private fun isConnected(): Boolean = connectivityManager.run {
-        getNetworkCapabilities(activeNetwork)?.run {
-            hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) && hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
-        }
-    } ?: false
-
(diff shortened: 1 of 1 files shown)


Elyniss (approved):

jakobuid:
Have you been able to test it manually at some point? So that the scenario that you described is reproducible on your device?

I'm asking because I haven't noticed any regression.

In the scenario that you describe when Wi-Fi is an active network and you disconnect from LTE it doesn't affect the connection because Wi-Fi is a primary network. No changes have been detected.

However, when in the above scenario you disconnect from Wi-Fi, the onLost callback is indeed triggered but since the primary network has been changed the onNetworkAvailable and onCapabilitiesChanged are triggered as well so that the connection is kept active and switched to LTE.

Can you verify if the following changes fix the Bluetooth issue on the Samsung device described in https://github.com/WalletConnect/WalletConnectKotlinV2/issues/556 ?

I'm not able to reproduce it.

jakubuid:
Hey @kamilargent I've released a SNAPSHOT build for you to check if the Bluetooth issue is still there:

```
implementation(platform("com.walletconnect:android-bom:1.4."))
implementation("com.walletconnect:android-core") 
implementation("com.walletconnect:sign")
```
 Please let me know the results of your testing

TalhaAli00 (approved):

kamilargent:
> Have you been able to test it manually at some point? So that the scenario that you described is reproducible on your device?
> 
> I'm asking because I haven't noticed any regression.
> 
> In the scenario that you describe when Wi-Fi is an active network and you disconnect from LTE it doesn't affect the connection because Wi-Fi is a primary network. No changes have been detected.
> 
> However, when in the above scenario you disconnect from Wi-Fi, the onLost callback is indeed triggered but since the primary network has been changed the onNetworkAvailable and onCapabilitiesChanged are triggered as well so that the connection is kept active and switched to LTE.
> 
> Can you verify if the following changes fix the Bluetooth issue on the Samsung device described in #556 ?
> 
> I'm not able to reproduce it.

The issue I am describing here happens on Asus Zenfone 8 Android 12 on the wallet connect example app. I tested this on this branch. In the described case: I get an `onLost` network callback for an LTE network - since we are subscribed to all valid network changes - not only the active network.

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/connection/ConnectivityState.kt
```diff
@@ -16,34 +16,31 @@ internal class ConnectivityState(context: Context) {
...
         }
     }
 
-    private fun updateIfStateChanged(network: Network) {
-        val isNowConnected = isNowConnected(network)
-
-        if (isNowConnected) {
+    fun setAvailability(isAvailable: Boolean, onAvailable: () -> Unit = {}, onLost: () -> Unit = {}) {
+        onLost()
+        if (isAvailable) {
```
Elyniss, line 34: Isn't it always true after once set to true?

jakubuid: This value is set every time when an onLost or onAvailable is triggered

onLost - we check whether the list of networks is not empty which defines this value
onAvailable - we check whether a given network satisfies defined capabilities. From Android documentation, we should not receive any other network with different capabilities BUT from what I've learned from testing this API on different devices I'm going to sleep better with this extra safe check

Elyniss: not valid. was confused from name shadowed variable

TalhaAli00 (approved):

kamilargent:
After the latest changes all of the issues on my end were resolved. All good from my end to merge it.

PR #582, opened 2023-01-14
Refactor/extract cacao updated

TalhaAli00:


Elyniss (approved):

jakobuid (approved):
q: Why is it needed to check the right order of params in the SDKSignature? Why can the order be different?

TalhaAli00:
The order matters for when we create a new object. Our parameters `t`, `s`, and `m` have to be in that order. To make it dynamic would add complexity that we won't really benefit from right now

PR #591, opened 2023-01-19
fix: reset shared prefs when InvalidProtocolBufferException

jakubuid:


Elyniss (approved):

androidCore/sdk/src/main/AndroidManifest.xml
```diff
@@ -4,4 +4,6 @@
 
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
     <uses-permission android:name="android.permission.INTERNET" />
+
+    <application android:allowBackup="true" android:fullBackupContent="@xml/backup_rules" />
```
TalhaAli00, line 8: why allow for backups?

sign/sdk/src/main/AndroidManifest.xml
```diff
@@ -2,5 +2,5 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.walletconnect.sign">
 
-    <application android:fullBackupContent="@xml/backup_rules" />
+    <application android:allowBackup="true" android:fullBackupContent="@xml/backup_rules" />
```
TalhaAli00, line 5: same comment, why enable backups?

androidCore/sdk/src/main/AndroidManifest.xml
jakubuid: Because based on the research it might help with the above exception on some exotic devices. It doesn't harm because we exclude what we care about which are all shared prefs, db and key files

sign/sdk/src/main/AndroidManifest.xml
jakubuid: same

androidCore/sdk/src/main/AndroidManifest.xml
jakubuid: Yea I found it somewhere in the google issue tracker. It's not like I have 100% guarantee that it helps I can agree that it might be restricted to some use-case that we are not aware of. I can remove it for now and if the issue still occurs we can think of adding it back

TalhaAli00 (approved):

PR #595, opened 2023-01-22
fix: workaround for error in tink lib

jakubuid:


androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreCryptoModule.kt
```diff
@@ -21,11 +21,12 @@ import java.io.File
 import java.security.KeyStore
 import com.walletconnect.android.internal.common.storage.KeyStore as WCKeyStore
 
+const val androidKeyStore = "AndroidKeyStore"
+const val sharedPrefsFile = "wc_key_store"
+const val keystoreAlias = "wc_keystore_key"
```
Elyniss, line 26: Since they are const vals could we make them in uppercase -> `ANDROID_KEY_STORE` etc.

Elyniss (approved):

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreCryptoModule.kt
TalhaAli00: and private?

androidCore/sdk/src/release/kotlin/com.walletconnect.android/di/CoreStorageModule.kt
```diff
@@ -102,7 +101,6 @@ private fun signingModule() = module {
             createSharedPreferences()
         } catch (e: Exception) {
             deleteSharedPreferences()
-            androidContext().databaseList().forEach { dbName -> deleteDBs(dbName) }
```
TalhaAli00, line 105: do we have to remove the deletion of storage from the storage module as well?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreCryptoModule.kt
jakubuid: 👍

androidCore/sdk/src/release/kotlin/com.walletconnect.android/di/CoreStorageModule.kt
jakubuid: Yes, I was able to repro this error in the storage module as well

TalhaAli00 (approved):

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
```diff
@@ -59,12 +59,23 @@ fun baseStorageModule() = module {
     single { JsonRpcHistory(get(), get()) }
 }
 
-object DBNames {
+
+object Database {
```
TalhaAli00, line 63: I would prefer a different name, maybe something like `DatabaseMetadata` or `DBUtil` something more descriptive

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
```diff
@@ -59,12 +59,23 @@ fun baseStorageModule() = module {
...
     const val ANDROID_CORE_DB_NAME = "WalletConnectAndroidCore.db"
+    const val SIGN_SDK_DB_NAME = "WalletConnectV2.db"
+    const val CHAT_SDK_DB_NAME = "WalletConnectV2_chat.db"
```
TalhaAli00, line 66: Also, why remove the db names out of the internal SDKs? We're now technically leaking the other SDKs knowledge into core

TalhaAli00: We could create a register function outside of the module and then build the `dbNames` list outside as well.

Elyniss (changes requested):
We shouldn't extract sdk databases into core

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
jakubuid: that works for me

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
jakubuid: because the encrypted shared prefs restart logic is placed inside the core, it happens before even reaching a specific sdk module. so to remove all dbs in this logic we should know the names of sdks dbs in core

Elyniss: True. But we could also remove databases containing some prefix i.e. `WalletConnect`

jakubuid: when this register function is supposed to be called? loading of crypto module happens without any knowledge of sdk specific module

Elyniss (approved):

TalhaAli00 (approved):

Elyniss (approved):

PR #603, opened 2023-01-25
Added user defined expiry logic and validation of user defined expiry

TalhaAli00:


Elyniss (approved):

sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -46,6 +44,7 @@ import com.walletconnect.utils.extractTimestamp
 import com.walletconnect.utils.isSequenceValid
 import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
+import java.util.*
 
 internal class SignEngine(
```
jakubuid, line 49: In Auth and Sign engines, we're missing the validation check when a user responds. In such a case we must return an error to the user and send the JsonRpcError to the other peer.

auth/sdk/src/main/kotlin/com/walletconnect/auth/engine/domain/AuthEngine.kt
```diff
@@ -118,6 +131,18 @@ internal class AuthEngine(
...
+                    try {
+                        withTimeout(irnParamsTtl.seconds) {
+                            jsonRpcInteractor.peerResponse
+                                .filter { response -> response.response.id == authRequest.id }
+                                .collect { cancel() }
+                        }
+                    } catch (e: TimeoutCancellationException) {
+                        _engineEvent.emit(SDKError(InternalError(e)))
+                    }
+                }
```
jakubuid, line 145: What kind of uses case this timeout covers?

auth/sdk/src/main/kotlin/com/walletconnect/auth/engine/domain/AuthEngine.kt
```diff
@@ -118,6 +131,18 @@ internal class AuthEngine(
 
                 pairingTopicToResponseTopicMap[pairingTopic] = responseTopic
                 onSuccess()
+
+                scope.launch {
+                    try {
+                        withTimeout(irnParamsTtl.seconds) {
```
jakubuid, line 137: Not sure if we should check against TTL here. TTL is for relay to know for how long a message gets to be stored. Expiry is set by a user to know for how long a request can be responded.

jakubuid: What happens when a phone call hits a user's phone?

auth/sdk/src/main/kotlin/com/walletconnect/auth/engine/domain/AuthEngine.kt
```diff
@@ -140,6 +165,13 @@ internal class AuthEngine(
         }
 
         val authParams: AuthParams.RequestParams = jsonRpcHistoryEntry.params
+
+        authParams.expiry?.let { expiry ->
+            if (CoreValidator.isExpiryNotWithinBounds(expiry)) {
+                return onFailure(InvalidExpiryException())
```
jakubuid, line 171: I cannot find a timeout for a respond method, is it omitted on purpose?

sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -306,9 +318,18 @@ internal class SignEngine(
...
+        if (!sessionStorageRepository.isSessionValid(topicWrapper)) {
+            return onFailure(CannotFindSequenceForTopic("$NO_SEQUENCE_FOR_TOPIC_MESSAGE$topic"))
         }
+
+        sessionStorageRepository.getSessionExpiryByTopic(topicWrapper)?.let { expiry ->
+            if (CoreValidator.isExpiryNotWithinBounds(expiry)) {
+                return onFailure(InvalidExpiryException())
+            }
+        } ?: return onFailure(CannotFindSequenceForTopic("$NO_SEQUENCE_FOR_TOPIC_MESSAGE$topic"))
+
```
jakubuid, line 332: Same question with timeout

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
```diff
@@ -205,7 +205,8 @@ object Sign {
             val sessionTopic: String,
             val method: String,
             val params: String,
-            val chainId: String
+            val chainId: String,
+            val expiry: Long? = null
```
jakubuid, line 209: We have to handle that in w3w as well

auth/sdk/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
```diff
@@ -100,6 +100,7 @@ object Auth {
             val statement: String?,
             val requestId: String?,
             val resources: List<String>?,
+            val expiry: Long? = null
```
jakubuid, line 103: We have to handle that in w3w as well

auth/sdk/src/main/kotlin/com/walletconnect/auth/engine/domain/AuthEngine.kt
TalhaAli00: I don't see why we would have a timeout function on the respond. With the CoreValidator returning false during this check on the expiry here, we're sure that the request isn't expired. If the request isn't expired, we are still able to respond to it. And once we publish the response, the expiry no longer matters for the wallet.

Let me know if my logic is correct or if I'm missing something

auth/sdk/src/main/kotlin/com/walletconnect/auth/engine/domain/AuthEngine.kt
TalhaAli00: I've added a new variable `requestTtlInSeconds` that will represent how long the timeout should be for the request instead of the relay TTL.

In the case of a phone call, the wallet could be closed and then the callback for the request function is skipped anyways. It's a similar scenario if the `onFailure` callback would be called but gets interrupted by a phone call.

sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
TalhaAli00: discussion happening in other comment

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
TalhaAli00: request is only for dapps

auth/sdk/src/main/kotlin/com/walletconnect/auth/client/Auth.kt
TalhaAli00: request is only for dapps

PR #624, opened 2023-02-07
Added nullable requestId to Sign.Params.Request to allow users to kno…

TalhaAli00:
…w what ID was sent

### DIFF
--- a/sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
+++ b/sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
@@ -206,7 +206,9 @@ object Sign {
             val method: String,
             val params: String,
             val chainId: String
-        ) : Params()
+        ) : Params() {
+            var requestId: Long? = null
+        }
 
         data class Update(
             val sessionTopic: String,
--- a/sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
+++ b/sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
@@ -56,6 +56,14 @@ object Sign {
             ) : Model()
         }
 
+        data class SentRequest(
+            val requestId: Long,
+            val sessionTopic: String,
+            val method: String,
+            val params: String,
+            val chainId: String
+        ) : Params()
+
         sealed class Namespace : Model() {
 
             data class Proposal(
@@ -206,9 +214,7 @@ object Sign {
             val method: String,
             val params: String,
             val chainId: String
-        ) : Params() {
-            var requestId: Long? = null
-        }
+        ) : Params()
 
         data class Update(
             val sessionTopic: String,
--- a/sign/sdk/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
+++ b/sign/sdk/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
@@ -144,15 +144,31 @@ class SignProtocol : SignInterface {
         }
     }
 
+    @Deprecated(
+        "The onSuccess callback has been replaced with a new callback that returns Sign.Model.SentRequest",
+        replaceWith = ReplaceWith("this.request(request, onSuccessWithSentRequest, onError)", "com.walletconnect.sign.client")
+    )
     @Throws(IllegalStateException::class)
-    override fun request(request: Sign.Params.Request, onSuccess: (Sign.Params.Request) -> Unit, onError: (Sign.Model.Error) -> Unit) {
+    override fun request(request: Sign.Params.Request, onSuccess: (Sign.Params.Request) -> Unit, onSuccessWithSentRequest: (Sign.Model.SentRequest) -> Unit, onError: (Sign.Model.Error) -> Unit) {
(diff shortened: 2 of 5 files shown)


sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
```diff
@@ -206,7 +206,9 @@ object Sign {
             val method: String,
             val params: String,
             val chainId: String
-        ) : Params()
+        ) : Params() {
+            var requestId: Long? = null
```
jakubuid, line 210: Why nullable?

sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -258,7 +258,7 @@ internal class SignEngine(
             })
     }
 
-    internal fun sessionRequest(request: EngineDO.Request, onSuccess: () -> Unit, onFailure: (Throwable) -> Unit) {
+    internal fun sessionRequest(request: EngineDO.Request, onSuccess: (Long) -> Unit, onFailure: (Throwable) -> Unit) {
```
jakubuid, line 261: Let's add it to the w3w as well

Elyniss (approved):

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
TalhaAli00: to not make this a breaking change. Changing the signature of this class or the onSuccess function will cause errors for existing developers

prashanOS: It might be clearer to create a new model - `SentRequest` with a `val requestId: Long`. 

Combining it here with this Params model is kind of confusing - if I as a consumer set `requestId` (since it is a `var`), I would expect it to be respected. But, as far as I can tell - SignEngine will still create a new ID and use that instead.

TalhaAli00: The issue with changing the signature to another class for the `onSuccess` callback again leads to a breaking change. I agree that a new type would be best in this situation but we also need to not create a bad UX for developers who have already implemented the SDK

sign/sdk/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
TalhaAli00: Can you clarify what you mean? W3W doesn't have the `sessionRequest` function

Elyniss (approved):

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -144,8 +144,12 @@ class SignProtocol : SignInterface {
         }
     }
 
+    @Deprecated(
+        "The onSuccess callback has been replaced with a new callback that returns Sign.Model.SentRequest",
+        replaceWith = ReplaceWith("this.request(request, onSuccessWithSentRequest, onError)", "com.walletconnect.sign.client")
+    )
     @Throws(IllegalStateException::class)
-    override fun request(request: Sign.Params.Request, onSuccess: (Sign.Params.Request) -> Unit, onError: (Sign.Model.Error) -> Unit) {
+    override fun request(request: Sign.Params.Request, onSuccess: (Sign.Params.Request) -> Unit, onSuccessWithSentRequest: (Sign.Model.SentRequest) -> Unit, onError: (Sign.Model.Error) -> Unit) {
```
jakubuid, line 152: @TalhaAli00 Do we need `onSuccessWithSentRequest: (Sign.Model.SentRequest) -> Unit` in that deprecated method?

jakubuid (approved):

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
TalhaAli00: it's the only way for the compiler to know the difference between the two functions. Without it, both functions conflict with each other

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -158,6 +162,20 @@ class SignProtocol : SignInterface {
         }
     }
 
+    @Throws(IllegalStateException::class)
+    override fun request(request: Sign.Params.Request, onSuccess: (Sign.Model.SentRequest) -> Unit, onError: (Sign.Model.Error) -> Unit) {
+        checkEngineInitialization()
+        try {
+            signEngine.sessionRequest(
+                request = request.toEngineDORequest(),
+                onSuccess = { requestId -> onSuccess(request.toSentRequest(requestId)) },
```
prashanOS, line 171: FYI - this technically breaks clients, since you're not calling the old callback anymore. Arguably, this is a bit more insidious because it won't be a compilation failure.

PR #645, opened 2023-02-13
feat: CAIP-25

jakubuid:


Elyniss:
Looks like migration check is failing

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
```diff
@@ -192,7 +180,9 @@ object Sign {
         ) : Params()
 
         data class Connect(
-            val namespaces: Map<String, Model.Namespace.Proposal>,
+            val namespaces: Map<String, Model.Namespace.Proposal>? = null,
```
TalhaAli00, line 183: does namespaces need to be nullable here? Can't it stay non-nullable as the old signature?

TalhaAli00 (approved):

sign/sdk/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
jakubuid: According to CAIP-25 it might be undefined

Elyniss (approved):

PR #660, opened 2023-02-16
fix: Fix faulty sha256 hashing

Elyniss:


TalhaAli00 (approved):

jakubuid (commented):
Why it's faulty?

Elyniss:
It was always meant for keys. I needed to use it to hash normal String as well. 

@jakubuid

jakubuid (approved):

PR #671, opened 2023-02-20
feat: batch subcribe

jakubuid:


Elyniss (approved):
Looks good, I wonder whether we should start calling batchSubscribe on AuthSDK, and ChatSDK in engine `setup()` calls

jakubuid:
> Looks good, I wonder whether we should start calling batchSubscribe on AuthSDK, and ChatSDK in engine `setup()` calls

It's already in AuthSDK, do you want me to add it in Chat as well?

Elyniss:
Must have missed AuthEngine change. 

>  do you want me to add it in Chat as well?

Yes 😄!

TalhaAli00 (approved):

PR #672, opened 2023-02-21
add generation of new JWT when old JWT is expired

TalhaAli00:


jakubuid (approved):

jakubuid (changes requested):
The first relay url gets to be overridden with the new one with a new jwt, otherwise the new connection always is initialized with the expired jwt

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -23,10 +26,15 @@ import retrofit2.Retrofit
 import retrofit2.converter.moshi.MoshiConverterFactory
 import java.util.concurrent.TimeUnit
 
+@Suppress("LocalVariableName")
 fun coreAndroidNetworkModule(serverUrl: String, jwt: String, connectionType: ConnectionType, sdkVersion: String) = module {
     val DEFAULT_BACKOFF_SECONDS = 5L
     val TIMEOUT_TIME = 5000L
 
+    single(named(AndroidCommonDITags.RELAY_URL)) {
```
Elyniss, line 34: Imo should be factory

Elyniss (approved):

jakubuid (approved):

PR #679, opened 2023-02-23
feat: Verify cacao on resolve identity

Elyniss:


jakubuid (approved):

chat/sdk/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
```diff
@@ -269,8 +271,12 @@ internal class ChatEngine(
...
         return runCatching { identitiesRepository.getAccountId(identityKey) }.onFailure {
             return resolveIdentityUseCase(identityKey).mapCatching { response ->
-                AccountId(decodeDidPkh(response.cacao.payload.iss)).also { accountId ->
-                    identitiesRepository.insertIdentity(identityKey, accountId)
+                if (CacaoVerifier(projectId).verify(response.cacao)) {
+                    AccountId(decodeDidPkh(response.cacao.payload.iss)).also { accountId ->
+                        identitiesRepository.insertIdentity(identityKey, accountId)
+                    }
+                } else {
+                    throw Throwable("Invalid cacao")
```
TalhaAli00, line 279: instead of throwing, can we emit an SDK error?

TalhaAli00 (approved):

chat/sdk/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
Elyniss: It's wrapped in result. Every place `resolveIdentity()` is called emits error or logs it

PR #681, opened 2023-02-23
feat: Disable inviting the same accounts

Elyniss:


chat/sdk/src/main/kotlin/com/walletconnect/chat/storage/InvitesStorageRepository.kt
```diff
@@ -21,6 +21,8 @@ internal class InvitesStorageRepository(private val invites: InvitesQueries) {
         }
     }
 
+    fun checkIfAccountsHaveExistingInvite(inviterAccount: String, inviteeAccount: String): Boolean = invites.checkIfAccountsHaveExistingInvite(inviterAccount, inviteeAccount).executeAsOne()
```
TalhaAli00, line 24: what do you think about changing this to `executeAsOneOrNull`? In case either account don't exist

TalhaAli00 (approved):

jakubuid (approved):

Issue #732, opened 2023-03-23
ACCOUNT_ADDRESS_REGEX does not comply with CAIP.


stKhaDgar:
**Describe the bug**
The address in the blockchain that we would like to support via **WalletConnect** contains a `:` 
_(e.g. 0:38d4c6ef92bd7ff4426eb232af3ff87bea525eff112f93eb28069e54bca89e4b)_, 
so your **[isAccountIdCAIP10Compliant](https://github.com/WalletConnect/WalletConnectKotlinV2/blob/ef6281c5b1218b03b199ddaca2207e63e1536285/androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/utils/CoreValidator.kt#L10)** function _(in the **CoreValidator** file)_ will not pass this address because the split method will return 4 lines instead of 3.

To avoid this problem, we replaced `:` with `%3A` _(encoding from the ASCII table)_, but then your regular will not resolve `%`:
`private const val ACCOUNT_ADDRESS_REGEX: String = "^([a-zA-Z0-9]{1,64})|([a-zA-Z0-9]{1,128})$"`
This code is from your **CoreValidator** file, here is a link to the line: **[CoreValidator:38](https://github.com/WalletConnect/WalletConnectKotlinV2/blob/ef6281c5b1218b03b199ddaca2207e63e1536285/androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/utils/CoreValidator.kt#L38)**

Can you please tell me what manual you used to write this regular? What I can see from the [CAIP](https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md#syntax) description in the **Syntax** section regarding `account_id`- I see that their regular supports the `%` symbol - `[-.%a-zA-Z0-9]{1,128}`.

Could you add support for this symbol?
Thanks in advance for your attention!

**SDK Version**
 - Client: Kotlin
 - Version android-core:1.11.1

Elyniss:
Hey! Thanks for the report. I believe we overlooked the regex change in Octoober 2022: https://github.com/ChainAgnostic/CAIPs/commit/222c1cd845827791416f3458930c7af80b968f48

stKhaDgar:
@Elyniss 
👋 I am very grateful for your reply. 
Is there any timeframe when you will be able to release this fix?

> Out of curiosity may I ask what blockchain is this?
[...]

PR #733, opened 2023-03-23
fix: Comply with CAIP10 regex

Elyniss:
Resolves: #732 

Casa specs: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-10.md#syntax


Elyniss:
Android Core tests fail due to issues with RPC. Will retry once rpc-proxy is back up

PR #742, opened 2023-03-27
✨ Make network client timeout customizable

mitsinsar:
Purpose of this PR is to make network client's timeout customizable. 

In our wallet, we support both WC V1 and WC V2. We use 10 seconds as timeout for V1. However for V2 it is 5 seconds as default and it makes things on our end a little bit inconsistent. 

Thanks in advance!

### DIFF
--- a/androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/model/NetworkClientTimeout.kt
+++ b/androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/model/NetworkClientTimeout.kt
@@ -1,0 +1,16 @@
+package com.walletconnect.android.internal.common.model
+
+import java.util.concurrent.TimeUnit
+
+data class NetworkClientTimeout(
+    val timeout: Long,
+    val timeUnit: TimeUnit
+) {
+
+    companion object {
+        fun getDefaultTimeout() = NetworkClientTimeout(
+            timeout = 5000L,
+            timeUnit = TimeUnit.MILLISECONDS
+        )
+    }
+}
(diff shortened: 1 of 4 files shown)


TalhaAli00 (approved):

jakubuid (approved):

jakubuid (changes requested):
Hi! Thanks for the PR. To make this change more robust I suggest adding a validation that the min value for time out is 5s, otherwise return an error. With this we'll avoid breaking the connection with timeout being too low.

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/model/NetworkClientTimeout.kt
```diff
@@ -0,0 +1,16 @@
+package com.walletconnect.android.internal.common.model
+
+import java.util.concurrent.TimeUnit
+
+data class NetworkClientTimeout(
```
jakubuid, line 5: Since this class is public it cannot be in `com.walletconnect.android.internal` package. Please move it to `com.walletconnect.android.relay`

TalhaAli00:
@jakubuid thinking about this more, we should probably put a min and max that is allowed. We don't want to potentially raise any issues with edge-case values

jakubuid:
> @jakubuid thinking about this more, we should probably put a min and max that is allowed. We don't want to potentially raise any issues with edge-case values. Specifically we probably don't want a timeout that is longer than the rebalancing timer.
> 
> Maybe we do a min of 5 seconds and a max of 1 min? 

Sounds good to me

Elyniss (approved):

jakubuid (approved):

mitsinsar:
Unfortunately, I am not able to resolve conflicts; 
`Only those with write access to this repository can merge pull requests.`
I'll be waiting for you to resolve them, then I can merge.

Thanks!

mitsinsar:
I assume I need your assistance for tests, @Elyniss 😇

Elyniss:
On it 🫡

Elyniss:
Imo, all good, @mitsinsar 

Android-Core:
Screenshot 2023-03-29 at 13 23 59

Relay: 
Screenshot 2023-03-29 at 13 24 48

mitsinsar:
Thanks for your quick response @Elyniss. Since I don't have permission to merge it, this pr will be waiting for your action. 

Thanks a lot!

PR #748, opened 2023-03-29
feat: change rpc id for client to server

jakubuid:


Elyniss (approved):

TalhaAli00 (approved):
What do you think about creating a `SDKRpc` interface that has an id field defaulting to the new id generation function? Then we would potentially guarantee all old and new rpc classes are using the same id function without having to call it every time

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -73,6 +73,7 @@ internal class JsonRpcInteractor(
         }
 
         val requestJson = serializer.serialize(payload) ?: return onFailure(IllegalStateException("JsonRpcInteractor: Unknown result params"))
+        println(":kobe: request: $requestJson")
```
Elyniss: Pls remove

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -102,6 +103,7 @@ internal class JsonRpcInteractor(
 
         val jsonResponseDO = response.toJsonRpcResponse()
         val responseJson = serializer.serialize(jsonResponseDO) ?: return onFailure(IllegalStateException("JsonRpcInteractor: Unknown result params"))
+        println(":kobe: response: $responseJson")
```
Elyniss: Pls remove

TalhaAli00 (approved):

PR #752, opened 2023-03-30
Add bitset logic to each SDK for the User-Agent

TalhaAli00:


androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -42,8 +43,19 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
 
     single(named(AndroidCommonDITags.INTERCEPTOR)) {
         Interceptor { chain ->
+            val binaryRepresentationWOLeadingZeros = getAll<BitSet>().let { listOfBitsets ->
```
jakubuid, line 46: What stands for WO?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/utils/UtilFunctions.kt
```diff
@@ -29,8 +30,11 @@ fun Expiry.isSequenceValid(): Boolean = seconds > CURRENT_TIME_IN_SECONDS
 val String.Companion.HexPrefix
     get() = "0x"
 
-fun <T : SerializableJsonRpc> Module.addSerializerEntry(value: KClass<T>): KoinDefinition<*> =
+fun <T : SerializableJsonRpc> Module.addSerializerEntry(value: KClass<T>): KoinDefinition<KClass<T>> =
     single(qualifier = named("key_${value.getFullName()}")) { value }
 
-fun Module.addDeserializerEntry(key: String, value: KClass<*>): KoinDefinition<*> =
+fun Module.addSdkBitset(key: String, bitsetWithOrder: BitSet): KoinDefinition<BitSet> =
+    single(qualifier = named("${key}_${bitsetWithOrder.length()}")) { bitsetWithOrder }
```
jakubuid, line 37: I wonder why we need a "key" in the qualifier. The length itself should be enough since the position of each bit is unique

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
Elyniss: WithOut?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -42,8 +43,19 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
...
+                val combinedBitset = listOfBitsets.reduce { acc, bitSet ->
+                    acc.or(bitSet)
+                    acc
+                }
+
+                combinedBitset.toByteArray().joinToString {
+                    "%8s".format(Integer.toBinaryString(it.toInt() and 0xFF)).replace(' ', '0')
+                }.replaceFirst("^0+(?!$)".toRegex(), "")
+            }
```
Elyniss, line 55: This code might need some tests

Elyniss (changes requested):
Overall code checks out, but I had to run it REPL. Some tests are necessary

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
TalhaAli00: It stands for without. I can add it in or maybe change the function name if it'll make it clearer

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/utils/UtilFunctions.kt
TalhaAli00: I'll have to double check to make sure the Bitset toString will be enough to have unique values. Maybe we just use the binary representation

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
TalhaAli00: this was generated by Copilot lol. I'll add a test

jakubuid:
> 

I see.

jakubuid:
> > When running sing.wallet I'm having 1001. Shouldn't be more like 100100?
> 
> No, the `1001` is correct. The old Sign Wallet uses Sign and Push. Sign is in the 2^0 position and Push is the 2^3 position. That's why you see both of those flags enabled

I'm asking whether we should include the trailing zeros. 1001 vs 100100? The function name says that no leading zeros are expected, what about trailing ones?

TalhaAli00:
> > > When running sing.wallet I'm having 1001. Shouldn't be more like 100100?
> > 
> > 
> > No, the `1001` is correct. The old Sign Wallet uses Sign and Push. Sign is in the 2^0 position and Push is the 2^3 position. That's why you see both of those flags enabled
> 
> I'm asking whether we should include the trailing zeros. 1001 vs 100100? The function name says that no leading zeros are expected, what about trailing ones?

Because we're using a little-endian ordered list, the array actually starts from the right. So `1001` is in the order of Push first then Sign when reading from left to right. The ordering is defaulted by the `toByteArray` function

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -42,8 +45,17 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
...
         Interceptor { chain ->
+            val sdkBitwiseFlags = getAll<BitSet>().let { listOfBitsets ->
+                val combinedBitset = listOfBitsets.reduce { acc, bitSet ->
+                    acc.or(bitSet)
+                    acc
+                }
+
+                combinedBitset.toBinaryString().removeLeadingZeros()
+            }
+
```
Elyniss, line 56: Could we extract this generation code into some utils that we would later have unit tests against? Currenlty unit tests test this code, but if it changes unit test wont catch any possible errors

TalhaAli00: That's fair. I'll add it to the tests

Elyniss (approved):

PR #756, opened 2023-04-04
fix: typo

jakubuid:


build.gradle.kts
```diff
@@ -39,6 +39,7 @@ subprojects {
             extensions.configure(BaseExtension::class.java) {
                 packagingOptions {
                     with(resources.excludes) {
+                        add("META-INF/versions/9/previous-compilation-data.bin")
```
kacperoak, line 42: This is a temporary coroutines bug fix. Personally, I would downgrade and remove it or add TODO to remove it when the coroutines version is updated (maybe in the Versions.kt next to coroutines version)

TalhaAli00: I would personally add the todo to this line instead of downgrading. The main reason for upgrading was to enable the ability for us to move towards creating a `CompletableFuture` for Java developers so we can move away from blocking methods.

The latest version of Coroutines comes with helper functions for converting Coroutines to CompleteableFuture.
https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jdk8/src/future/Future.kt

TalhaAli00 (approved):

PR #764, opened 2023-04-05
feat: Add Sync API interface

Elyniss:
Next in line: #765

### DIFF
--- a/sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
+++ b/sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
@@ -1,7 +1,37 @@
 package com.walletconnect.sync.client
 
+import com.walletconnect.android.internal.common.model.AccountId
+
 interface SyncInterface {
-    interface SyncDelegate {}
+    interface SyncDelegate {
+        fun onSyncUpdate(onSyncUpdate: Sync.Events.OnSyncUpdate)
+
+        fun onConnectionStateChange(state: Sync.Model.ConnectionState)
+
+        fun onError(error: Sync.Model.Error)
+    }
+
     fun setSyncDelegate(delegate: SyncDelegate)
+
     fun initialize(params: Sync.Params.Init)
+
+    /**
+     * Caution: This function is blocking and runs on the current thread.
+     * It is advised that this function be called from background operation
+     */
+    fun getMessage(params: Sync.Params.GetMessage): Result<String> // discuss: Result or nullable String
+
+    fun register(params: Sync.Params.Register, onSuccess: () -> Unit, onError: (Sync.Model.Error) -> Unit)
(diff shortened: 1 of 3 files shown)


sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
```diff
@@ -1,7 +1,37 @@
...
+
+    fun set(params: Sync.Params.Set, onSuccess: () -> Unit, onError: (Sync.Model.Error) -> Unit)
+
+    fun delete(params: Sync.Params.Delete, onSuccess: () -> Unit, onError: (Sync.Model.Error) -> Unit)
+
+    /**
+     * Caution: This function is blocking and runs on the current thread.
+     * It is advised that this function be called from background operation
+     */
+    fun getStores(params: Sync.Params.GetStores) : Result<Sync.Type.StoreMap> // discuss: Result or nullable String
```
Elyniss, line 36: What are your thoughts?

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
```diff
@@ -1,7 +1,37 @@
...
     fun setSyncDelegate(delegate: SyncDelegate)
+
     fun initialize(params: Sync.Params.Init)
+
+    /**
+     * Caution: This function is blocking and runs on the current thread.
+     * It is advised that this function be called from background operation
+     */
+    fun getMessage(params: Sync.Params.GetMessage): Result<String> // discuss: Result or nullable String
+
```
Elyniss, line 23: What are your thoughts?

jakubuid: From quick research handling Result from Java projects might be problematic, since this class is public and  in theory might be used by other projects I'd stick to nullable string

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
jakubuid: Same

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
Elyniss: Changed to nullable

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncInterface.kt
Elyniss: Changed to nullable

jakubuid (approved):

TalhaAli00 (approved):

PR #773, opened 2023-04-07
feat: Sync engine use-cases skeleton

Elyniss:


Elyniss:
Realized I never added any `onSucces` and `onFailure` to usecases. Will add in following PR that resolves TODO/implements usecases

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncProtocol.kt
```diff
@@ -85,4 +103,6 @@ class SyncProtocol : SyncInterface {
             wrapWithRunCatching(onError) { block() }
         }
     }
+
+    private fun validateAccountId(accountId: AccountId) = if (!accountId.isValid()) throw InvalidAccountIdException(accountId) else Unit
```
TalhaAli00, line 107: Since this is business logic, I think it should be in the engine and called on every function

sync/sdk/src/main/kotlin/com/walletconnect/sync/common/model/StoreState.kt
```diff
@@ -0,0 +1,3 @@
+package com.walletconnect.sync.common.model
+
+internal interface StoreState : Map<String, String>
```
TalhaAli00, line 3: Can we use value classes instead of primitives?

TalhaAli00 (approved):

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncProtocol.kt
Elyniss: The way I look at it is param validation and could rightfully be here. If I were to put it in every use-case then some would be validated multiple times (i.e Register which calls GetMessage)

sync/sdk/src/main/kotlin/com/walletconnect/sync/common/model/StoreState.kt
Elyniss: Sooo, you mean
```suggestion
internal interface StoreState : Map<StoreKey, StoreValue>
```

Elyniss: Address in following PR #779 with commit: https://github.com/WalletConnect/WalletConnectKotlinV2/pull/779/commits/817f47d40930f84643bd2e417feb7b0768c335c6

sync/sdk/src/main/kotlin/com/walletconnect/sync/client/SyncProtocol.kt
Elyniss: Address in following PR #779 with commit: https://github.com/WalletConnect/WalletConnectKotlinV2/pull/779/commits/817f47d40930f84643bd2e417feb7b0768c335c6

PR #785, opened 2023-04-10
feat: Extract signature and message verification from CacaoVerifier

Elyniss:
Independent PR, no need to sync with Sync API PR queue


androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/signing/SignatureVerifier.kt
```diff
@@ -0,0 +1,15 @@
+package com.walletconnect.android.internal.common.signing
+
+import com.walletconnect.android.cacao.signature.SignatureType
+import com.walletconnect.android.internal.common.model.ProjectId
+import com.walletconnect.android.internal.common.signing.eip1271.EIP1271Verifier
+import com.walletconnect.android.internal.common.signing.eip191.EIP191Verifier
+import com.walletconnect.android.internal.common.signing.signature.Signature
+
+abstract class SignatureVerifier(private val projectId: ProjectId) {
```
TalhaAli00, line 9: Does this need to be an abstract class? Can we simply this into just a function?

Elyniss: Could be but I wanted to have an object with project id dependency in koin

Elyniss: I will tinker with just function to see how it looks ;)

Elyniss: Check out the [`44c867e` (#785)](https://github.com/WalletConnect/WalletConnectKotlinV2/pull/785/commits/44c867e83c2d061bbf8519c918c4f820b2c40780)

TalhaAli00 (approved):

PR #814, opened 2023-04-13
Create initial screen for new dapp

kacperoak:
resolves 777


Elyniss (approved):
LGTM! Remember to merge WCTopAppBar from develop that has version of BOM

samples/dapp/src/main/kotlin/com/walletconnect/sample/dapp/domain/model/Chain.kt
```diff
@@ -0,0 +1,17 @@
...
+    val chainReference: String,
+    val icon: Int,
+    val color: String,
+    val methods: List<String>,
+    val events: List<String>,
+    val order: Int,
+)
+
+fun Chains.toChain() =
+    Chain(chainName, chainNamespace, chainReference, icon, color, methods, events, order)
```
TalhaAli00, line 17: why do we need the `Chain` class if we have the `Chains` class?

jakubuid (approved):

samples/dapp/src/main/kotlin/com/walletconnect/sample/dapp/domain/model/Chain.kt
kacperoak: You're right. In this sample, this model brings nothing. I will delete it

PR #837, opened 2023-04-24
wip: how to listen to sync events in other protocols

Elyniss:
Draft as it's not ready but showcases the idea on how to integrate, one way, sync -> chat interaction

Main takeaways: 
- Chat engine listens for sync protocol updates/events on specific stores. Whenever they happen a particular chat use-case is invoked as if it received the event from the chat client call

What's missing:
- Integration when chat client calls needs to invoke sync updates

### DIFF
--- a/chat/sdk/src/main/kotlin/com/walletconnect/chat/engine/sync/updates/OnChatInviteUpdateUseCase.kt
+++ b/chat/sdk/src/main/kotlin/com/walletconnect/chat/engine/sync/updates/OnChatInviteUpdateUseCase.kt
@@ -1,0 +1,19 @@
+package com.walletconnect.chat.engine.sync.updates
+
+import com.walletconnect.android.sync.common.model.Events
+import com.walletconnect.android.sync.common.model.SyncUpdate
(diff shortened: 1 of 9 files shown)


chat/sdk/src/main/kotlin/com/walletconnect/chat/engine/sync/updates/OnChatInviteUpdateUseCase.kt
```diff
@@ -0,0 +1,19 @@
+package com.walletconnect.chat.engine.sync.updates
+
+import com.walletconnect.android.sync.common.model.Events
+import com.walletconnect.android.sync.common.model.SyncUpdate
+import com.walletconnect.chat.engine.use_case.calls.AcceptInviteUseCase
+
+internal class OnChatInviteUpdateUseCase(
+    private val acceptInviteUseCase: AcceptInviteUseCase,
+) {
+    operator fun invoke(event: Events.OnSyncUpdate) {
```
TalhaAli00, line 10: I don't think we should be using the client-facing models in our domain layer. Can we create a EngineDO type for this?

TalhaAli00 (approved):

PR #838, opened 2023-04-24
Create yml files for samples releases

kacperoak:


.github/workflows/ci_release_sample_dapp.yml
```diff
@@ -0,0 +1,39 @@
+name: SAMPLES RELEASE
+
+on:
+  push:
+    branches: ["release"]
```
kacperoak, line 5: This is something we can discuss about when we want to trigger these actions

kacperoak:
After sync changing it into draft and try to move it in gradle plugin: 
https://firebase.google.com/docs/app-distribution/android/distribute-gradle?hl=en&authuser=0&apptype=apk

.github/workflows/ci_release_sample.yml
```diff
@@ -0,0 +1,40 @@
+name: Sample releases
+
+on:
+  push:
+#    branches: ["master"]
```
kacperoak, line 5: lets talk about trigger

.github/workflows/ci_release_sample_dapp.yml
jakubuid: I think we should have a stage between master and develop. I'd say staging that triggers sending samples to testers via firebase. Is this what you meant?

kacperoak: Yes

.github/workflows/ci_release_sample.yml
Elyniss, line 5:
```suggestion
on: workflow_dispatch
```

TalhaAli00 (approved):

jakubuid (approved):

PR #860, opened 2023-05-09
Fix UnsafisfiedLinkError with SqlCipher

kacperoak:


TalhaAli00 (approved):

kacperoak:
> Are able to check if this fixes issue reported by OpenSea? Have we been able to repro?

I tried but couldn't reproduce it

PR #862, opened 2023-05-10
Add support to xml views for web3modal

### LINKED ISSUE #854: Support Navigation Component via XML

kacperoak:
Resolves #854


web3/modal/src/main/kotlin/com/walletconnect/web3/modal/ui/Web3ModalSheet.kt
```diff
@@ -0,0 +1,57 @@
...
+            setContent {
+                Web3ModalComposeView()
+            }
+        }
+    }
+
+    @Composable
+    private fun Web3ModalComposeView() {
+        val navController = rememberAnimatedNavController()
+        (dialog as? ComponentDialog)?.onBackPressedDispatcher?.addCallback(
```
kacperoak, line 38: it's a bit of a trick. Because dialog/dialogfragment don't have onbackdispatcher available by default. Without it, our compose dialog dismiss itself every time someone uses native back. In the case of fragment everything is ok, but we have to provide this view to the user as a Dialog

Elyniss (approved):

jakubuid (approved):

jakubuid (approved):

PR #886, opened 2023-05-23
fix: unusable master key

jakubuid:
Based on the GitHub discussion it might be helpful to delete the master key first, before deleting encrypted shared prefs and DB.

Source: https://github.com/google/tink/issues/535#issuecomment-912170221


Elyniss (approved):
We don't have a way to test it, right?

jakubuid:
> We don't have a way to test it, right?

Correct, I verified that it doesn't break anything.
Regarding verification it's observation drive development

TalhaAli00 (approved):

PR #951, opened 2023-06-26
Feat/core/fix integration workflow

Elyniss:


jakubuid (approved):

.github/workflows/ci_instrumented_tests.yml
```diff
@@ -64,46 +70,46 @@
...
-#        run: ./gradlew --stop
+
+      - name: Create AVD and generate snapshot for caching
+        if: steps.avd-cache.outputs.cache-hit != 'true'
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: ${{ matrix.api-level }}
+          target: ${{ matrix.target }}
+          arch: ${{ matrix.arch }}
+          profile: ${{ matrix.profile }}
```
TalhaAli00, line 81: if we add more targets, profiles, and architectures, will the matrix run all combinations?

Elyniss: yes

Elyniss: there is also a way to add single configuration

TalhaAli00 (approved):

.github/workflows/ci_instrumented_tests.yml
TalhaAli00: I think it's better to be able to run all possibilities. it will take longer to complete the tests but we'll have better coverage

PR #962, opened 2023-06-29
fix: handle nullpointerexception when dealing with a storage

jakubuid:


Elyniss (approved):

androidCore/sdk/src/main/kotlin/com/walletconnect/android/pairing/engine/domain/PairingEngine.kt
```diff
@@ -299,21 +300,25 @@ internal class PairingEngine(
...
-            return@let pairing.isValid()
+            return@let pairing.isExpired()
         }
     }
 
     private fun generateTopic(): Topic = Topic(randomBytes(32).bytesToHex())
 
-    private fun Pairing.isValid(): Boolean = (expiry.seconds > CURRENT_TIME_IN_SECONDS).also { isPairingValid ->
-        if (!isPairingValid) {
+    private fun Pairing.isExpired(): Boolean = (expiry.seconds > CURRENT_TIME_IN_SECONDS).also { isValid ->
```
TalhaAli00, line 309: Based on the name, true means the function is expired but I don't think that is what we want to convey. Maybe change it to `isNotExpired` to match the intention when true.

Also, why the name change?

jakubuid: I agree to the method name confusion - fixed.

Also, why the name change? - We already have a method named the same way, so it was confusing. This particular one check expiry so the change name seems natural

kacperoak (approved):

TalhaAli00 (approved):

PR #960, opened 2023-06-28
Create WalletConnect module and ModalCore module

kacperoak:
- Create two modules: `WalletConnectModal` module and `ModalCore` module
- Migrate our sample dapp to use `WalletConnectModal`
- Move explorer api logic to `AndroidCore`

Test flow: 
Run sample dapp and connect to wallets

PR #968, opened 2023-06-30
Create modal state 

kacperoak:
This PR is based on #960


TalhaAli00 (approved):

Elyniss (approved):

product/modal/src/main/kotlin/com/walletconnect/modal/ui/state/ModalState.kt
```diff
@@ -0,0 +1,42 @@
...
+    ModalState(navController)
+}
+
+class ModalState(
+    private val navController: NavController
+) {
+    val isOpen: Boolean
+        get() = navController.currentDestination.isModal()
+
+    val isOpenState: State<Boolean>
```
jakubuid, line 28: Why do we need isOpenState - wouldn't be enough to expose only isOpenFlow?

kacperoak: Ye good point. Lets move this responsibility to the developer

PR #985, opened 2023-07-05
fix: add MasterKey key generation parameter spec into CryptoModule

jakubuid:


Elyniss:
Why is this necessary?

jakubuid:
> Why is this necessary?

To unify MasterKey generation on Crypto and Storage modules and eliminate possibly faulty creation on some of the Samsung devices

TalhaAli00 (approved):

TalhaAli00:
Was there an issue reported that led to this? Or did you read something somewhere?

Elyniss (approved):

PR #1002, opened 2023-07-11
feat: implements failover logic to relay.org

jakubuid:


androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -21,20 +22,36 @@ import com.walletconnect.utils.Empty
...
 @JvmSynthetic
 fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType, sdkVersion: String, timeout: NetworkClientTimeout? = null) = module {
     val DEFAULT_BACKOFF_SECONDS = 5L
-
     val networkClientTimeout = timeout ?: NetworkClientTimeout.getDefaultTimeout()
+    var SERVER_URL: String = serverUrl
+    var wasFailOvered = false
+
+    factory(named(AndroidCommonDITags.RELAY_URL)) {
+        val jwt = wcKoinApp.koin.get<GenerateJwtStoreClientIdUseCase>().invoke(SERVER_URL)
```
Elyniss, line 47:
```suggestion
        val jwt = get<GenerateJwtStoreClientIdUseCase>().invoke(SERVER_URL)
```

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -21,20 +22,36 @@ import com.walletconnect.utils.Empty
...
+        Uri.parse(SERVER_URL)
+            .buildUpon()
+            .appendQueryParameter("auth", jwt)
+            .appendQueryParameter("ua", wcKoinApp.koin.get(named(AndroidCommonDITags.USER_AGENT)))
```
Elyniss, line 51:
```suggestion
            .appendQueryParameter("ua", get(named(AndroidCommonDITags.USER_AGENT)))
```

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
Elyniss: We mustn't use global wcKoinApp. It most definitely will break Instrumented tests

Elyniss (changes requested):

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
jakubuid: Good catch - some leftovers from testing different approaches

androidCore/sdk/src/main/kotlin/com/walletconnect/android/relay/NetworkClientTimeout.kt
```diff
@@ -20,7 +20,7 @@ data class NetworkClientTimeout(
 
     companion object {
 
-        private const val MIN_TIMEOUT_LIMIT_AS_MILLIS = 5_000L
```
Elyniss, line 23: Why is this necessary? This was added by a contributor and possibly could break their production

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -68,19 +75,45 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
...
+            try {
+                if (wasFailOvered && request.url.host == DEFAULT_RELAY_URL) {
+                    chain.proceed(request.newBuilder().url(get<String>(named(AndroidCommonDITags.RELAY_URL))).build())
+                } else {
+                    chain.proceed(request)
+                }
+            } catch (e: Exception) {
+                if (request.url.host == DEFAULT_RELAY_URL) {
+                    SERVER_URL = "$FAIL_OVER_RELAY_URL?projectId=${Uri.parse(SERVER_URL).getQueryParameter("projectId")}"
+                    wasFailOvered = true
```
kacperoak, line 90: This flag stays true forever?

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -68,19 +75,45 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
...
-                response.request.run {
-                    if (Uri.parse(serverUrl).host == this.url.host) {
-                        val relayUrl = get<String>(named(AndroidCommonDITags.RELAY_URL))
-                        this.newBuilder().url(relayUrl).build()
-                    } else {
-                        null
-                    }
-                }
-            })
+            .addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY))
```
TalhaAli00, line 113: double check to see if these logs are still sent to Logcat in a release build. We should not be logging network traffic outside of debug

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -68,19 +75,45 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
...
+    single(named(AndroidCommonDITags.FAIL_OVER_INTERCEPTOR)) {
+        Interceptor { chain ->
+            val request = chain.request()
+            try {
+                if (wasFailOvered && request.url.host == DEFAULT_RELAY_URL) {
+                    chain.proceed(request.newBuilder().url(get<String>(named(AndroidCommonDITags.RELAY_URL))).build())
+                } else {
+                    chain.proceed(request)
+                }
+            } catch (e: Exception) {
```
TalhaAli00, line 87: what kind of exception is expected to be thrown? We should more specific about when we want to use the fail-over url instead defaulting to the failover in case of an error

androidCore/sdk/src/main/kotlin/com/walletconnect/android/relay/NetworkClientTimeout.kt
jakubuid: It's necessary because 5s timeout is to short for the failover logic to be executed without unnecessary additional tries

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
jakubuid: Not forever - the init value is false. It's set to true only if the failover logic happens. It's not set back to false because Scarlet instance is a singleton so we cannot re-init it with different url it means that whenever the connection is restarted we need to intercept and replace the failing url with the new one to reduce initialization latency.

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
jakubuid: Yeah I agree - overlooked this when cleaning up before creating a PR
I'm gonna add logic to keep this interceptor for debug builds - it's helpful to see what's going on in the network while debugging

androidCore/sdk/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
jakubuid: I was not able to get consistency in one error - I can add the set of errors that occurs

androidCore/sdk/src/main/kotlin/com/walletconnect/android/relay/NetworkClientTimeout.kt
jakubuid: The solution work fine with 5s timeout as well, so there's no need to increase it

Elyniss (changes requested):

Elyniss (approved):

TalhaAli00 (approved):

PR #1027, opened 2023-07-18
feat: Fix url so that already subscribed dapps are not visible in W3I…

Elyniss:
… explorer

### DIFF
--- a/protocol/push/src/main/kotlin/com/walletconnect/push/wallet/engine/PushWalletEngine.kt
+++ b/protocol/push/src/main/kotlin/com/walletconnect/push/wallet/engine/PushWalletEngine.kt
@@ -190,16 +190,16 @@ internal class PushWalletEngine(
                 val dappMetaData: AppMetaData = withContext(Dispatchers.IO) {
                     // Fetch dapp metadata from explorer api
                     val listOfDappHomepages = runCatching {
-                        explorerRepository.getAllDapps().listings.associateBy { listing -> listing.homepage.host }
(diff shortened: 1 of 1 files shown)


protocol/push/src/main/kotlin/com/walletconnect/push/wallet/engine/PushWalletEngine.kt
```diff
@@ -190,16 +190,16 @@ internal class PushWalletEngine(
...
 
                     // Find dapp metadata for dapp uri
-                    val (dappHomepageHost: String?, dappListing: Listing) = listOfDappHomepages.entries.filter { (dappHomepageHost, dappListing) ->
-                        dappHomepageHost != null && dappListing.description != null
-                    }.firstOrNull { (dappHomepageHost, _) ->
-                        dappHomepageHost != null && dappHomepageHost.contains(dappUri.host!!)
+                    val (dappHomepageUri: Uri, dappListing: Listing) = listOfDappHomepages.entries.filter { (_, dappListing) ->
+                        dappListing.description != null
+                    }.firstOrNull { (dappHomepageUri, _) ->
+                        dappHomepageUri.toString().contains(dappUri.toString())
```
TalhaAli00, line 202: why switch from comparing just the host to the full uri contents?

Elyniss: Right, might not be necessary. Changed back to host comparing

TalhaAli00 (approved):

jakubuid (approved):

PR #1028, opened 2023-07-18
feat: Reduce sync requests

Elyniss:


core/android/src/main/kotlin/com/walletconnect/android/history/ReduceSyncRequestsUseCase.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Reduction algorithm:
+     * 1. Decrypt all messages, discard any exceptions
+     * 2. Serialize messages to know which method it belongs to
+     * 3. !Temporary! Remove messages duplicates by json rpc id
+     * 4. Split them by method into sets. Method: wc_syncDelete, wc_syncSet and remaining
+     * 5. Reduce any wc_syncDelete along with corresponding wc_syncSet requests
+     * 6. Retain order by retaining initial message only if is within remaining wc_syncDelete, wc_syncSet and remaining requests
+     */
```
TalhaAli00, line 28: each of these steps should be a separate private function for easier readability. Also, for more idiomatic Kotlin, you could make each function an extension function of List/Set and use functions like `map`, `filterNotNull`, ect.

i.e.
```kotlin
fun List<HistoryMessage>.decryptMessages(): MutableList<Triple<ClientJsonRpc, HistoryMessage, String>> = this.map{}.filterNotNull()
```

core/android/src/main/kotlin/com/walletconnect/android/history/ReduceSyncRequestsUseCase.kt
```diff
@@ -0,0 +1,77 @@
...
+    suspend operator fun invoke(historyMessages: List<HistoryMessage>) {
```
TalhaAli00, line 29: Also, because you're using the `Triple`, I don't know if you're able to benefit from the usage of sets. I would try it out with lists first instead of converting to a set multiple times and see if ti changes anything

jakubuid (approved):

TalhaAli00 (approved):

PR #1039, opened 2023-07-21
Add redirect on hold screen

### LINKED ISSUE #1020: On hold screen

kacperoak:
Resolves #1020


jakubuid (approved):

product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/ui/navigation/ModalNavGraph.kt
```diff
@@ -40,20 +42,29 @@ internal fun ModalNavGraph(
...
+        ) { backStackEntry ->
+            val wallet = state.wallets.find { it.id == backStackEntry.arguments?.getString(Route.OnHold.walletIdKey, String.Empty) }!!
+            RedirectOnHoldScreen(navController = navController, uri = state.uri, wallet = wallet)
+        }
     }
 }
 
 internal fun NavGraphBuilder.animatedComposable(
     route: String,
+    arguments: List<NamedNavArgument> = emptyList(),
```
TalhaAli00, line 62: It looks like this is always set so is the default value valid?

TalhaAli00 (commented):
Android does not have universal links, just deep links. We should be able to remove all references of the universal links from the codebase

product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/ui/navigation/ModalNavGraph.kt
kacperoak: Its not needed anymore. I moved animation to NavHost and use default `composable()` extension

Elyniss (approved):

PR #1045, opened 2023-07-24
[Do not merge] Add Instrumented tests to CI

Elyniss:


Elyniss:
Example gradle task name on GHA:
Screenshot 2023-07-24 at 16 38 32

When there is no $GITHUB_RUNNER_ID variable task will be named:
Screenshot 2023-07-24 at 16 40 01

.github/workflows/ci_instrumented_tests.yml
```diff
@@ -11,13 +11,13 @@ on:
...
-#   push:
-#     branches:
-#       - develop
-#   pull_request:
-#     branches:
-#       - develop
-#       - master
+  push:
+    branches:
+      - develop
```
TalhaAli00, line 16: correct me if I'm wrong, but this will cause tests to run when pushing directly to develop right? I don't think we need this since the tests would not prevent the commit

TalhaAli00 (approved):

jakubuid (approved):

PR #1061, opened 2023-07-31
[WCM] Move paring into WalletConnectModalViewModel.kt

kacperoak:
- Move `Core.Pairing` inside WCM  viewModel and fix retry 
- Improve retry UI 
- migrate W3I and Sample Dapp with new navigation


product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/ui/WalletConnectModal.kt
```diff
@@ -97,20 +108,32 @@ private fun LoadingModalState() {
...
+@Composable
+private fun ErrorModalState(closeModal: () -> Unit) {
+    Column(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(300.dp),
+        verticalArrangement = Arrangement.Center,
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text(text = "Something goes wrong", style = TextStyle(color = ModalTheme.colors.onBackgroundColor, fontSize = 18.sp))
```
TalhaAli00, line 120:
```suggestion
        Text(text = "Something went wrong", style = TextStyle(color = ModalTheme.colors.onBackgroundColor, fontSize = 18.sp))
```

product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/ui/WalletConnectModalViewModel.kt
```diff
@@ -1,90 +1,113 @@
...
-        WalletConnectModalDelegate.wcEventModels.map { event ->
-            when (event) {
-                is Modal.Model.ApprovedSession -> WalletConnectModalEvents.SessionApproved
-                is Modal.Model.RejectedSession -> WalletConnectModalEvents.SessionRejected
-                else -> WalletConnectModalEvents.NoAction
-            }
-        }.onEach { event ->
-            _modalEvents.emit(event)
+    private fun initModalState() {
+        val sessionParams = WalletConnectModal.sessionParams ?: throw IllegalStateException("Session params missing")
```
TalhaAli00, line 44: I think we need to re-think the logic here.
Since Sessionparams is optional and if someone doesn't provide the SessionParams, then this function will throw an exception. 
But if the developer didn't provide the SessionParams by default, then we shouldn't show them an error message

maybe instead of the try-catch, we do something like `WalletConnectModal.sessionParams?.let {`

kacperoak: what would be the point of not adding these parameters specifically? The modal will then enter an error state and there is nothing the user can do. I can move it inside try catch, then log this error

TalhaAli00: then let's not make then nullable and fail earlier

TalhaAli00 (approved):

product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/client/Modal.kt
```diff
@@ -36,6 +37,12 @@ object Modal {
             val chainId: String,
             val expiry: Long? = null
         ) : Params()
+
+        data class SessionParams(
```
jakubuid, line 41: IMO better name for it would be `InitParams`

jakubuid (approved):

PR #1087, opened 2023-08-10
Improve eip1271 signature validation

jakubuid:


Elyniss (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/signing/eip1271/EIP1271Verifier.kt
```diff
@@ -64,9 +65,14 @@ internal object EIP1271Verifier {
         val id = generateId()
         val request: Request = Request.Builder().url(projectId.prefixWithRpcUrl()).post(createBody(address, data, id)).build()
         val response: Response = OkHttpClient().newCall(request).execute()
+        val responseString = response.body?.string() ?: throw Exception("Response body is null")
+        val validResponseResult = getResponseResult(getValidResponse(id))
+        val responseResult = getResponseResult(responseString)
```
TalhaAli00, line 70: With the complexity growing, what do you think about replacing OkHttpClient and manual parsing with Retrofit+Moshi?

TalhaAli00 (approved):

kacperoak (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/signing/eip1271/EIP1271Verifier.kt
jakubuid: For now it's good enough, I'd add those once we need to add some other stuff

PR #1091, opened 2023-08-11
Check expiry before adding requests to queue

jakubuid:


protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -647,11 +646,14 @@ internal class SignEngine(
         getPendingSessionRequests()
             .map { pendingRequest -> pendingRequest.toSessionRequest(metadataStorageRepository.getByTopicAndType(pendingRequest.topic, AppMetaDataType.PEER)) }
             .map { sessionRequest ->
-                scope.launch {
-                    supervisorScope {
-                        val verifyContext = verifyContextStorageRepository.get(sessionRequest.request.id) ?: VerifyContext(sessionRequest.request.id, String.Empty, Validation.UNKNOWN, String.Empty)
-                        val sessionRequestEvent = EngineDO.SessionRequestEvent(sessionRequest, verifyContext.toEngineDO())
-                        sessionRequestsQueue.addLast(sessionRequestEvent)
+                if (CoreValidator.isExpiryWithinBounds(sessionRequest.expiry)) {
```
kacperoak, line 649: Any log in else case? Do we need that information?

kacperoak (approved):

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -647,11 +646,14 @@ internal class SignEngine(
         getPendingSessionRequests()
             .map { pendingRequest -> pendingRequest.toSessionRequest(metadataStorageRepository.getByTopicAndType(pendingRequest.topic, AppMetaDataType.PEER)) }
             .map { sessionRequest ->
```
TalhaAli00, line 648: instead of `.map`, `.forEach` would better explain what you want to do

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/model/EngineDO.kt
```diff
@@ -76,6 +76,7 @@ internal sealed class EngineDO {
         val chainId: String?,
         val peerAppMetaData: AppMetaData?,
         val request: JSONRPCRequest,
+        val expiry: Expiry?
```
TalhaAli00, line 79: should this be defaulted to null?

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
jakubuid: I think it's unnecessary because it's on init and we don't need information that given request is expired - it wouldn't be added to the queue

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/model/EngineDO.kt
jakubuid: there is no need because null can come from dto object

TalhaAli00 (approved):

PR #1102, opened 2023-08-21
Add Redirect Screen. Refactor GetAWalletRoute.kt and HelpRoute.kt

kacperoak:


product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/internal/Web3ModalStates.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.walletconnect.web3.modal.ui.components.internal.commons.button.ButtonSize
+import com.walletconnect.web3.modal.ui.components.internal.commons.button.TryAgainButton
+import com.walletconnect.web3.modal.ui.theme.Web3ModalTheme
+
+@Composable
+internal fun LoadingModalState() {
+    Box(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(300.dp),
```
TalhaAli00, line 24: should we define a ratio ratio instead of doing a fixed height? This will make weird boxes on wider screens

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/internal/Web3ModalStates.kt
```diff
@@ -0,0 +1,46 @@
...
+        LoadingSpinner()
+    }
+}
+
+@Composable
+internal fun ErrorModalState(retry: () -> Unit) {
+    Column(
+        modifier = Modifier
+            .fillMaxWidth()
+            .height(300.dp),
```
TalhaAli00, line 36: same

TalhaAli00 (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/Web3ModalViewModel.kt
```diff
@@ -39,6 +39,21 @@ internal class Web3ModalViewModel : ViewModel() {
...
+    internal fun retryConnection(onSuccess: () -> Unit) {
+        try {
+            val sessionParams = Web3Modal.sessionParams
+            val connectParams = Modal.Params.Connect(
+                sessionParams.requiredNamespaces,
+                sessionParams.optionalNamespaces,
+                sessionParams.properties,
+                pairing
+            )
+            Web3Modal.connect(connectParams, onSuccess) { Timber.e(it.throwable) }
```
jakubuid, line 51: Should we call handleError in the callback?

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/internal/commons/inputs/SearchInput.kt
```diff
@@ -0,0 +1,125 @@
...
+                            onSearchValueChange(String.Empty)
+                            onClearClick.invoke()
+                        }
+                    }
+                }
+                HorizontalSpacer(width = 12.dp)
+            }
+        },
+    )
+//    }
```
jakubuid, line 115: Remove

jakubuid (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/internal/Web3ModalStates.kt
kacperoak: I don't quite understand your idea. At this point we have a fixed-height initial state with the loader. This fixed-height state transforms into a content-dependent state. Aspect ratio here won't mean much because later and we can't define a fixed or aspect ratio as component size.

This size may change later as I try to start supporting horizontal orientation on the root box

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/Web3ModalViewModel.kt
kacperoak: I think not in this case. In the feature that will be replaced with custom snackbar. Right now it WIP

PR #1128, opened 2023-09-10
Feature/meta/add codeowners

TalhaAli00:


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @WalletConnect/kotlin-team
```
Elyniss, line 1: I think we should be more granular and add all users individually

kacperoak: if there is this group, is it 1 approval or each one separately?

TalhaAli00: I've added each of the members of the group as well.

Elyniss (approved):

PR #1129, opened 2023-09-11
Split inbox sample features for internal and external builds

TalhaAli00:


Elyniss (changes requested):
Could we close the PR. Sample will be removed anyway

PR #1147, opened 2023-09-19
[W3W] feat: new designs for verify api 

jakubuid:


core/android/src/main/core/migration/3.sqm
```diff
@@ -0,0 +1,4 @@
+
+-- migration from 3.db to 4.db
+
+ALTER TABLE VerifyContext ADD COLUMN is_scam INTEGER AS Boolean;
```
Elyniss, line 4: There will be conflicts. Let's decide which PR needs to merged first. This or https://github.com/WalletConnect/WalletConnectKotlinV2/pull/1145

Elyniss (changes requested):

core/android/src/main/core/migration/3.sqm
TalhaAli00: let's merge this first before #1145

core/android/src/main/core/com/walletconnect/android/sdk/storage/data/dao/VerifyContext.sq
```diff
@@ -4,20 +4,21 @@ CREATE TABLE VerifyContext (
     id INTEGER PRIMARY KEY NOT NULL,
     origin TEXT NOT NULL,
     validation TEXT AS Validation NOT NULL,
-    verify_url TEXT NOT NULL
+    verify_url TEXT NOT NULL,
+    is_scam INTEGER AS Boolean
```
TalhaAli00, line 8: should this be nullable? Can we make it `NOT NULL` and default to false?

core/android/src/main/core/migration/3.sqm
TalhaAli00: Going along with the comment above, can we default to false instead?

core/android/src/main/kotlin/com/walletconnect/android/verify/data/model/AttestationResult.kt
```diff
@@ -0,0 +1,6 @@
+package com.walletconnect.android.verify.data.model
+
+data class AttestationResult(
+    val origin: String,
+    val isScam: Boolean?
```
TalhaAli00, line 5: similar to the comment on the table, can we default this to false?

TalhaAli00 (approved):

core/android/src/main/kotlin/com/walletconnect/android/verify/data/model/AttestationResult.kt
Elyniss: same as above

core/android/src/main/core/migration/3.sqm
Elyniss: #1145 got changed to not have conflicts here

jakubuid: I'd not default it to false because it would mean that for all defaulted cases all dapps are legit which is not true. it's better to have it as null

core/android/src/main/core/com/walletconnect/android/sdk/storage/data/dao/VerifyContext.sq
jakubuid: Exactly, it's better to keep default value as null

core/android/src/main/kotlin/com/walletconnect/android/verify/data/model/AttestationResult.kt
jakubuid: Same

PR #1153, opened 2023-09-20
[Notify] Add Integration Test and improve workflow

Elyniss:
Requires update of dev env vars, namely:
- `PROD_GM_PROJECT_ID`
- `PROD_GM_SECRET`


protocol/notify/build.gradle.kts
```diff
@@ -24,6 +24,13 @@ android {
         }
 
         buildConfigField(type = "String", name = "SDK_VERSION", value = "\"${requireNotNull(extra.get(KEY_PUBLISH_VERSION))}\"")
+        buildConfigField("String", "PROJECT_ID", "\"${System.getenv("WC_CLOUD_PROJECT_ID") ?: ""}\"")
+        buildConfigField("String", "PROD_GM_PROJECT_ID", "\"${System.getenv("PROD_GM_PROJECT_ID") ?: ""}\"")
+        buildConfigField("String", "PROD_GM_SECRET", "\"${System.getenv("PROD_GM_SECRET") ?: ""}\"")
```
TalhaAli00, line 29: since these are needed to test, instead of defaulting to empty strings, let's use `requreNotNull` for each of the env vars. 
That way if they are missing, it won't silently fail

protocol/notify/src/androidTest/kotlin/com/walletconnect/notify/di/OverrideModule.kt
```diff
@@ -0,0 +1,34 @@
...
+private const val SHARED_PREFS_FILE = "wc_key_store"
+private const val KEY_STORE_ALIAS = "wc_keystore_key"
+
+// When called more that once, different `storagePrefix` must be defined.
+@JvmSynthetic
+internal fun overrideModule(
+    relay: RelayConnectionInterface,
+    pairing: PairingInterface,
+    pairingController: PairingControllerInterface,
+    storagePrefix: String = "test_"
```
TalhaAli00, line 21: if you don't default it, then it will force the prefix to be passed each time it's used instead of having to read the comment

Elyniss: True, will change this in other files as well, thanks!

Elyniss:
I will research how to properly add env vars to GHA

jakubuid (approved):

kacperoak (approved):

PR #1154, opened 2023-09-20
[Notify] Polish before hackaton

Elyniss:
Went through code one last time, and found some missing pieces

- Extract Notify Server Url from usecases
- Add validation plan. I don't want to force it now becaues it requires adding new table for storing keys from did.json files

### DIFF
--- a/protocol/notify/src/main/kotlin/com/walletconnect/notify/common/NotifyServerUrl.kt
+++ b/protocol/notify/src/main/kotlin/com/walletconnect/notify/common/NotifyServerUrl.kt
@@ -3,7 +3,7 @@ package com.walletconnect.notify.common
 import androidx.core.net.toUri
 
 @JvmInline
-internal value class NotifyServerUrl(val url: String) {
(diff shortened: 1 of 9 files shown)


protocol/notify/src/main/kotlin/com/walletconnect/notify/common/NotifyServerUrl.kt
```diff
@@ -0,0 +1,10 @@
+package com.walletconnect.notify.common
+
+import androidx.core.net.toUri
+
+@JvmInline
+internal value class NotifyServerUrl(val url: String) {
```
TalhaAli00, line 6: maybe we default the url to the const?

TalhaAli00 (approved):

protocol/notify/src/main/kotlin/com/walletconnect/notify/common/NotifyServerUrl.kt
Elyniss: Yeah we could

PR #1165, opened 2023-09-25
[W3M] Expose modal Web3ModalComponent and ComponentDelegate

kacperoak:
- Expose Web3ModalComponent for developers
- Add component delegate

### DIFF
--- a/product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/ComponentDelegate.kt
+++ b/product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/ComponentDelegate.kt
@@ -1,0 +1,38 @@
+package com.walletconnect.web3.modal.ui.components
+
+import com.walletconnect.web3.modal.client.Web3Modal
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.onEach
+
+internal sealed class ComponentEvent {
+    object ModalHiddenEvent: ComponentEvent()
+    object ModalExpandedEvent: ComponentEvent()
+}
+
+internal object ComponentDelegate {
+
+    val modalComponentEvent: MutableSharedFlow<ComponentEvent> = MutableSharedFlow()
+
+    var isModalOpen: Boolean = false
+
+    fun setDelegate(delegate: Web3Modal.ComponentDelegate) {
+        modalComponentEvent.onEach {
+            when(it) {
+                ComponentEvent.ModalHiddenEvent -> delegate.onModalHidden().also { isModalOpen = false }
+                ComponentEvent.ModalExpandedEvent -> delegate.onModalExpanded().also { isModalOpen = true }
+            }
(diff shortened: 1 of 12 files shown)


product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/ComponentDelegate.kt
```diff
@@ -0,0 +1,38 @@
+package com.walletconnect.web3.modal.ui.components
+
+import com.walletconnect.web3.modal.client.Web3Modal
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.onEach
+
+internal sealed class ComponentEvent {
+    object ModalHiddenEvent: ComponentEvent()
+    object ModalExpandedEvent: ComponentEvent()
+}
```
TalhaAli00, line 10: In the Web3ModalState file, you check each of these values and return a boolean. 

You could also do
```suggestion
internal sealed class ComponentEvent(isOpen: Boolean) {
    object ModalHiddenEvent: ComponentEvent(false)
    object ModalExpandedEvent: ComponentEvent(true)
}
```

Then in Web3ModalSate where you listen to the event, you just grab the `isOpen` variable

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/button/Web3ModalState.kt
```diff
@@ -41,12 +39,20 @@ class Web3ModalState(
...
     private val getSelectedChainUseCase: GetSelectedChainUseCase = wcKoinApp.koin.get()
 
-    val isOpen = navController.currentBackStackEntryFlow.mapLatest { it.destination.route?.startsWith(Route.WEB3MODAL.path) ?: false }
+    val isOpen = ComponentDelegate.modalComponentEvent
+        .map { event ->
+            when (event) {
+                ComponentEvent.ModalHiddenEvent -> false
+                ComponentEvent.ModalExpandedEvent -> true
+            }
+        }
```
TalhaAli00, line 54: If you want to use the suggestion above, then this becomes

```suggestion
    val isOpen = ComponentDelegate.modalComponentEvent
        .map { event -> event.isOpen }
```

TalhaAli00 (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/ComponentDelegate.kt
jakubuid: Why would we need it? I think just by an event itself it's clear on which state modal is

jakubuid: Ah ok I see below

jakubuid (approved):

Issue #1170, opened 2023-10-02
CannotFindSequenceForTopic exception


vcoolish:
Getting lots of crashes on app
```
Exception com.walletconnect.android.internal.common.exception.CannotFindSequenceForTopic:
  at com.walletconnect.sign.engine.use_case.calls.RespondSessionRequestUseCase$respondSessionRequest$2.invokeSuspend (RespondSessionRequestUseCase.kt:49)
  at com.walletconnect.sign.engine.use_case.calls.RespondSessionRequestUseCase$respondSessionRequest$2.invoke (RespondSessionRequestUseCase.kt:2)
  at com.walletconnect.sign.engine.use_case.calls.RespondSessionRequestUseCase$respondSessionRequest$2.invoke (RespondSessionRequestUseCase.kt:1)
  at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn (Undispatched.kt:78)
  at kotlinx.coroutines.SupervisorKt.supervisorScope (Supervisor.kt:61)
  at com.walletconnect.sign.engine.use_case.calls.RespondSessionRequestUseCase.respondSessionRequest (RespondSessionRequestUseCase.kt)
  at com.walletconnect.sign.engine.domain.SignEngine.respondSessionRequest (SignEngine.kt)
  at com.walletconnect.sign.client.SignProtocol$respond$1.invokeSuspend (SignProtocol.kt:213)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:108)
  at kotlinx.coroutines.internal.LimitedDispatcher$Worker.run (LimitedDispatcher.kt:115)
  at kotlinx.coroutines.scheduling.TaskImpl.run (Tasks.kt:103)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.kt:584)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask (CoroutineScheduler.kt:793)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker (CoroutineScheduler.kt:697)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt)
```

**SDK Version**
 - Client: Kotlin
 - Version bom 1.16.0

Issue #1130, opened 2023-09-12
[Re-open] Getting older sessionRequest and data even if send new request


nguyennh-0786:
**Describe the bug**
Refer https://github.com/WalletConnect/WalletConnectKotlinV2/issues/1105 . The current issue has been closed, so I created a new one.

**SDK Version**
 - Client: Kotlin
 - Version: BOM 1.16.0

**To Reproduce**
Refer https://github.com/WalletConnect/WalletConnectKotlinV2/issues/1105 .

muratogat:
I also encounter the same problem and this is a real issue. 

I'm trying to use Uniswap on mobile browser through deeplinking.

After going through the SessionProposal, SessionApprove and onSessionSettleResponse Result flow, I return to Uniswap to initiate the actual transaction, which triggers another deeplink into my wallet.

The requestID and sessionTopic in the deeplink have nothing to with the actual topic and id in the request received over the connection afterwards. The one in the deeplink is correct, the other one in the actual SessionRequest is wrong / very old.

Trying to approve or reject this request leads to a crash in the SDK since a session with the topic in the sessionRequest actually does not exist.

I think better handling of the session cache or a way to trigger a cleanup is needed. Deleting and reinstalling the app would most likely solve the problem.

Using Web3Wallet, problem exists at least in 1.16 and also 1.17.

jakubuid:
Hey @muratogat thanks for bringing this up:
- Can you tell whether this issue happens also on the newly created session between wallet and dapp?
- Can you send the SessionRequest ID that is old/wrong?

muratogat:
Had the session with Uniswap with topic 0013c53d677832cc56bb8615842ec5843bfda407ad8f2e7c1a166173b4db3630 and closed is successfully (SessionDisconnect onSuccess).
Uniswap registers this and switches back to disconnected state.
Established a new session with Uniswap with topic e8aff8ae....f70621427. Session is settled.
Initiated a SessionRequest over this connection, trying to wrap ETH to WETH.
Deeplink is triggered. Request ID is 1696232947226047 , topic is e8aff8ae....f70621427 , same as above.

The session request ID is 1695815440135721 and the topic is 5d6aa3ca5525f305d79acc058d24adfb66c8a4a04a96ed6830d9b99a548dd947
[...]

Issue #1105, opened 2023-08-23
Getting older sessionRequest and data even if send new request


aadityapaliwal:
**Describe the bug**
I am getting older sessionRequest data on onSessionRequest callback and that too multiple time. Even after I disconnect and connect with new QR code, it's sending older data.

**SDK Version**
 - implementation("com.walletconnect:android-core")
 -  implementation("com.walletconnect:web3wallet")
 - implementation(platform("com.walletconnect:android-bom:1.15.0"))

**To Reproduce**
1. Pair a dapp with your wallet
2. It will start sending getNetwork, getMirrorNetwork, call events even if you didn't requested for.
3. Send a new request from your dapp with a different accountid
4. Now onSessionRequest, I am getting older request though.

**Expected behavior**
It should update new request and data.

**Screenshots**

**Device (please complete the following information):**
 - Device: Motorola G5
 - OS: Android 8.1.0
 - Browser [e.g. stock browser, chrome]

**Additional context**
This is the older request and data.
```
This is request I am getting ->

SessionRequest(topic=fde7b0aab07605ef2bd219e56c9e4084558a667c6f95629c4c768b6f71b7f9cd, chainId=hedera:296, peerMetaData=AppMetaData(name=Blade Wallet Demo App, description=, url=https://127.0.0.1:3001, icons=[], redirect=, verifyUrl=null), request=JSONRPCRequest(id=1692698541921573, method=call, params={"accountId":"0.0.622648","executable":"Sg4KAhAAEggIABAAGLiAJg==","isTransaction":false}))

This is the dapp with which I am currently paired with. You can check the topic is different with above sessionRequest ->

[...]

PR #1172, opened 2023-10-03
Bugfix/sign/respond session request catch error

TalhaAli00:


protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/DisconnectSessionUseCase.kt
```diff
@@ -18,11 +18,11 @@ import kotlinx.coroutines.supervisorScope
 internal class DisconnectSessionUseCase(
     private val jsonRpcInteractor: JsonRpcInteractorInterface,
     private val sessionStorageRepository: SessionStorageRepository,
-    private val logger: Logger
+    private val logger: Logger,
 ) : DisconnectSessionUseCaseInterface {
     override suspend fun disconnect(topic: String, onSuccess: () -> Unit, onFailure: (Throwable) -> Unit) = supervisorScope {
         if (!sessionStorageRepository.isSessionValid(Topic(topic))) {
-            throw CannotFindSequenceForTopic("$NO_SEQUENCE_FOR_TOPIC_MESSAGE$topic")
+            return@supervisorScope onFailure(CannotFindSequenceForTopic("$NO_SEQUENCE_FOR_TOPIC_MESSAGE$topic"))
```
jakubuid, line 25: What's the reason of this change?

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/ExtendSessionUseCase.kt
```diff
@@ -19,23 +19,23 @@ import com.walletconnect.sign.common.model.vo.clientsync.session.params.SignPara
...
         }
 
         val session = sessionStorageRepository.getSessionWithoutMetadataByTopic(Topic(topic))
         if (!session.isSelfController) {
-            throw UnauthorizedPeerException(UNAUTHORIZED_EXTEND_MESSAGE)
+            return@supervisorScope onFailure(UnauthorizedPeerException(UNAUTHORIZED_EXTEND_MESSAGE))
         }
         if (!session.isAcknowledged) {
-            throw NotSettledSessionException("$SESSION_IS_NOT_ACKNOWLEDGED_MESSAGE$topic")
+            return@supervisorScope onFailure(NotSettledSessionException("$SESSION_IS_NOT_ACKNOWLEDGED_MESSAGE$topic"))
```
jakubuid, line 38: The same question

jakubuid (commented):
What's the purpose of this change? Exceptions thrown in use cases should be already caught in SignProtocol class and exposed in onError callback to devs

kacperoak (approved):

TalhaAli00:
> What's the purpose of this change? Exceptions thrown in use cases should be already caught in SignProtocol class and exposed in onError callback to devs

This is to fix the issue Trust is seeing in their app. They are currently seeing around 2k crashes being reported and the stacktrace was shared in #1170. 

The protocol class doesn't capture the exceptions that are thrown and it causing crashes at runtime. I'm not 100% but my guess is that because we're launching our usecases in a supervisor scope, that is why the exceptions aren't being propagated back to the protocol class.

jakubuid:
> > What's the purpose of this change? Exceptions thrown in use cases should be already caught in SignProtocol class and exposed in onError callback to devs
> 
> This is to fix the issue Trust is seeing in their app. They are currently seeing around 2k crashes being reported and the stacktrace was shared in #1170.
> 
> The protocol class doesn't capture the exceptions that are thrown and it causing crashes at runtime. I'm not 100% but my guess is that because we're launching our usecases in a supervisor scope, that is why the exceptions aren't being propagated back to the protocol class.

Hmm that's interesting. Have you been able to reproduce this behavior? In my testing, I have not been able to.

Still, to me, it looks like a way of mitigating the issue not solving the root cause. Do we know why Trust is seeing `CannotFindSequenceForTopic` exception? Do we know how to repro?

I did my investigation: https://github.com/WalletConnect/WalletConnectKotlinV2/pull/1175
Based on:
https://github.com/WalletConnect/WalletConnectKotlinV2/issues/1105
https://github.com/WalletConnect/WalletConnectKotlinV2/issues/1130

I believe these might be connected

muratogat:
Perhaps I'm a bit a bit out of place here, but I just wanted to confirm that :
- IMO #1105 is the exact reason Trust or any other wallets are experiencing `CannotFindSequenceForTopic `
- I also don't see any of the exceptions being propageted to onError and lead to a crash instead. Even wrapping the block for approving-rejecting in try-catch doesn't work in that case. That's why I manually check if the topic is present before attempting any actions, as described [here](https://github.com/WalletConnect/WalletConnectKotlinV2/issues/1130#issuecomment-1742563971).

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/DisconnectSessionUseCase.kt
TalhaAli00: this will propagate the error as before it was just being thrown

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/ExtendSessionUseCase.kt
TalhaAli00: same as above

protocol/sign/src/test/kotlin/com/walletconnect/sign/engine/use_case/calls/DisconnectSessionUseCaseTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val disconnectSessionUseCase = DisconnectSessionUseCase(jsonRpcInteractor, sessionStorageRepository, logger)
+
+    @Before
+    fun setUp() {
+        every { logger.error(any() as String) } answers { }
+        every { logger.error(any() as Exception) } answers { }
+    }
+
+    @After
+    fun tearDown() {
```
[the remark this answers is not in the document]
TalhaAli00: removed

jakubuid (approved):

PR #1175, opened 2023-10-03
Fix: receiving pending requests on disconnected session

jakubuid:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/json_rpc/domain/JsonRpcInteractor.kt
```diff
@@ -55,7 +55,7 @@ internal class JsonRpcInteractor(
 
     override val isConnectionAvailable: StateFlow<Boolean> get() = relay.isConnectionAvailable
 
-    private val subscriptions: MutableMap<String, String> = mutableMapOf()
+    private var subscriptions: MutableMap<String, String> = mutableMapOf()
```
TalhaAli00, line 58: why do we need a var of a mutable map?

kacperoak (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/json_rpc/domain/JsonRpcInteractor.kt
jakubuid: To perform this operation on the list:

`  subscriptions = subscriptions.plus(topics.zip(acknowledgement.result).toMap()).toMutableMap()`

Elyniss (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/json_rpc/domain/JsonRpcInteractor.kt
TalhaAli00: if the problem before was that we were never updating the old map, then instead of using `plus` before, we can change to `plusAssign`. That should be equal to what you have here without having a mutable variable

TalhaAli00: Kotin docs: https://kotlinlang.org/docs/map-operations.html#add-and-update-entries

jakubuid: Nice, it works the same, thanks for showing this

PR #1182, opened 2023-10-06
feat: allow dapp extending a session

jakubuid:


protocol/sign/src/androidTest/kotlin/com/walletconnect/sign/test/client/SignClientInstrumentedAndroidTest.kt
```diff
@@ -233,6 +234,26 @@ class SignClientInstrumentedAndroidTest {
         launch(walletDelegate, dappDelegate)
     }
 
+    @Test
+    fun extendSessionByDapp() {
+        Timber.d("receiveSessionExtendByDapp: start")
```
kacperoak, line 239: Why we need logs in tests?

kacperoak (approved):

kacperoak (approved):

protocol/sign/src/androidTest/kotlin/com/walletconnect/sign/test/client/SignClientInstrumentedAndroidTest.kt
jakubuid: To see if it starts when sth goes wrong

TalhaAli00 (approved):

PR #1189, opened 2023-10-12
[Notify] Update statement gracefully

Elyniss:
We had to rephrase the statement, which means user has to sign new cacao. Old identity keys have to be removed from Keys Server and register with updated cacao. Updated db table to add tracking to signed payload and whether it user's or not


sample/wallet/build.gradle.kts
```diff
@@ -16,7 +16,7 @@ android {
         applicationId = "com.walletconnect.sample.wallet"
         minSdk = MIN_SDK
         targetSdk = TARGET_SDK
-        versionCode = SAMPLE_VERSION_CODE
+        versionCode = SAMPLE_VERSION_CODE + 1
```
Elyniss, line 19:
```suggestion
        versionCode = SAMPLE_VERSION_CODE
```

core/android/src/main/core/migration/5.sqm
```diff
@@ -0,0 +1,13 @@
+-- migration from 5.db to 6.db
+
+-- Up until now, we've been using the Identities table to cache the remote identities, hence we can drop it
+
+DROP TABLE IF EXISTS Identities;
```
TalhaAli00, line 5: should we be dropping the table? I think we can just do `ALTER TABLE` and add the new columns since they have defaults 
https://www.techonthenet.com/sqlite/tables/alter_table.php#:~:text=The%20syntax%20to%20ADD%20A,table_name

TalhaAli00 (approved):

core/android/src/main/core/com/walletconnect/android/sdk/storage/data/dao/Identities.sq
```diff
@@ -1,14 +1,25 @@
 CREATE TABLE Identities (
     id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
-    identity TEXT NOT NULL,
-    accountId TEXT NOT NULL
+    identity TEXT UNIQUE NOT NULL,
+    accountId TEXT NOT NULL,
+    cacao_payload TEXT DEFAULT NULL,
+    is_mine INTEGER AS Boolean DEFAULT 0
```
jakubuid, line 6: What do you mean by `is_mine`?

core/android/src/main/kotlin/com/walletconnect/android/keyserver/domain/IdentitiesInteractor.kt
```diff
@@ -44,17 +45,44 @@ class IdentitiesInteractor(
 ) {
     fun getIdentityKeyPair(accountId: AccountId): Pair<PublicKey, PrivateKey> = keyManagementRepository.getKeyPair(getIdentityPublicKey(accountId))
 
-    suspend fun registerIdentity(accountId: AccountId, statement: String, domain: String, resources: List<String>, onSign: (String) -> Cacao.Signature?): Result<PublicKey> = try {
+    suspend fun registerIdentity(accountId: AccountId, statement: String, domain: String, resources: List<String>, keyserverUrl: String, onSign: (String) -> Cacao.Signature?): Result<PublicKey> =
+        getAlreadyRegisteredIdentity(accountId, statement, domain, resources)
+            .recoverCatching { exception ->
+                when (exception) {
+                    is MissingKeyException -> handleNotYetGeneratedIdentities(accountId, statement, domain, resources, onSign).getOrThrow()
+                    is AccountHasNoCacaoPayloadStored, is AccountHasDifferentStatementStored -> handleIdentitiesOutdatedStatements(accountId, statement, domain, resources, keyserverUrl, onSign).getOrThrow()
```
jakubuid, line 53: What is a statement?

core/android/src/main/kotlin/com/walletconnect/android/keyserver/domain/IdentitiesInteractor.kt
```diff
@@ -44,17 +45,44 @@ class IdentitiesInteractor(
...
+        return runCatching {
+            val storedPublicKey = getIdentityPublicKey(accountId)
+            val cacaoPayload = identitiesRepository.getCacaoPayloadByIdentity(storedPublicKey.keyAsHex) ?: throw AccountHasNoCacaoPayloadStored(accountId)
+            val generatedPayload = generatePayload(accountId, storedPublicKey, statement, domain, resources).getOrThrow()
+            if (cacaoPayload.statement != generatedPayload.statement) throw AccountHasDifferentStatementStored(accountId)
+            storedPublicKey
+        }
+    }
+
+    private suspend fun handleNotYetGeneratedIdentities(accountId: AccountId, statement: String, domain: String, resources: List<String>, onSign: (String) -> Cacao.Signature?): Result<PublicKey> {
```
jakubuid, line 69: `handleNotYetGeneratedIdentities` - how can we handle identities that are not generated?

core/android/src/main/kotlin/com/walletconnect/android/keyserver/domain/IdentitiesInteractor.kt
```diff
@@ -75,13 +103,15 @@ class IdentitiesInteractor(
     private suspend fun resolveAndStoreIdentityRemotely(identityKey: String) = resolveIdentityUseCase(identityKey).mapCatching { response ->
         if (!CacaoVerifier(projectId).verify(response.cacao)) throw InvalidIdentityCacao()
         val accountId = AccountId(decodeDidPkh(response.cacao.payload.iss))
-        identitiesRepository.insertIdentity(identityKey, accountId)
+        identitiesRepository.insertIdentity(identityKey, accountId, response.cacao.payload, isMine = false)
```
jakubuid, line 106: `isMine` doesn't say whose identity. do we have an actor that we can use to have a better name?

core/android/src/main/core/com/walletconnect/android/sdk/storage/data/dao/Identities.sq
Elyniss: Previously this table was only used for storing other account Identities, with this feature we are saving users account as well so we could see their most recent cacao_payload signed and stored for given Identity

core/android/src/main/kotlin/com/walletconnect/android/keyserver/domain/IdentitiesInteractor.kt
Elyniss: It's part of SIWE/CAIP-122 message that is signed by the user. Specs -> https://github.com/ChainAgnostic/namespaces/blob/main/solana/caip122.md

core/android/src/main/kotlin/com/walletconnect/android/keyserver/domain/IdentitiesInteractor.kt
Elyniss: @jakubuid Do you think `generateIdentity` would be a better name?

core/android/src/main/kotlin/com/walletconnect/android/keyserver/domain/IdentitiesInteractor.kt
Elyniss: It allows to differentiate Identities by flagging which one I created, and which one was created by other account.

core/android/src/main/core/migration/5.sqm
Elyniss: I added UNIQUE Constraint which makes ALTER TABLE not an option. Alternatively, I could make a migration. However, I believe we could easily drop this table as this was only used for caching

core/android/src/main/core/com/walletconnect/android/sdk/storage/data/dao/Identities.sq
jakubuid: what do you think about `is_owner `?

Elyniss: Sounds good

PR #1191, opened 2023-10-13
Feature/update pairing isactive property

TalhaAli00:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/Pairing.kt
```diff
@@ -12,17 +12,17 @@ data class Pairing(
     val relayProtocol: String,
     val relayData: String?,
     val uri: String,
-    val isActive: Boolean,
     val registeredMethods: String
 ) : Sequence {
+    val isActive: Boolean
+        get() = (expiry.seconds > CURRENT_TIME_IN_SECONDS)
```
jakubuid, line 18: I'm not sure if we want to do this. Can you check such flow:
- open QR code on a dapp
- scan with a wallet
- dismiss proposal dialog on the wallet (do not approve nor reject)
- scan the same qr code
- is proposal dialog shown?

TalhaAli00: @jakubuid the wallet is showing that the pairing already exists. Is that expected?

jakubuid: We have regression with this change, I tested the flow above and users cannot scan the same QR code twice

TalhaAli00: Updated the logic and the above test case works now

jakubuid (approved):

Elyniss (approved):

kacperoak (approved):

PR #1195, opened 2023-10-16
[Notify/W3W Sample] feat: Notification name as channel id

Elyniss:
Screenshot 2023-10-16 at 10 47 31

Since we change how to manage notification types from human readable string to uuids we need to change channel names as well


jakubuid (approved):

protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/calls/GetNotificationTypesUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.walletconnect.android.internal.common.explorer.domain.usecase.GetNotifyConfigUseCase
+import com.walletconnect.notify.common.model.NotificationType
+import com.walletconnect.notify.data.storage.SubscriptionRepository
+
+internal class GetNotificationTypesUseCase(
+    private val getNotifyConfigUseCase: GetNotifyConfigUseCase,
+
+    ) : GetNotificationTypesUseCaseInterface {
+    override suspend fun getNotificationTypes(domain: String): Map<String, NotificationType> {
+        val notifyConfig = getNotifyConfigUseCase(domain).getOrThrow()
```
TalhaAli00, line 14: instead of throwing, we should just return the exception back to the developer. Or add the Throws annotation on protocol's `getNotificationTypes`

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/WalletFirebaseMessagingService.kt
```diff
@@ -37,6 +39,23 @@ class WalletFirebaseMessagingService : NotifyMessageService() {
...
+                val appDomain = URI(appUrl).host ?: throw IllegalStateException("Unable to parse domain from $appUrl")
+
+                NotifyClient.getNotificationTypes(Notify.Params.NotificationTypes(appDomain))[message.type]?.name
+                    ?: throw IllegalStateException("No notification type for topic: ${message.topic} and type: ${message.type}")
+            }.getOrElse {
+                Timber.e(it)
+                message.type
+            }
+
+        }
```
TalhaAli00, line 57: maybe we move this logic into the service? Then we can expose methods like `getChannelId` and `getChannelName`

protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/calls/GetNotificationTypesUseCase.kt
```diff
@@ -0,0 +1,29 @@
...
+
+        return notifyConfig.types.map { notificationType ->
```
TalhaAli00, line 16: you can just use `.associate` instead of `.map` with `toMap()`

TalhaAli00 (approved):

kacperoak (approved):

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/WalletFirebaseMessagingService.kt
Elyniss: I think this might need broader discussion, it will also be used in other SDKs. Not sure if those method are applicable to Sign or Auth SDKs

protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/calls/GetNotificationTypesUseCase.kt
Elyniss: Added annotation

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/WalletFirebaseMessagingService.kt
Elyniss: We decided to tackle this in next release

PR #1198, opened 2023-10-16
[Notify] feat: Change RegisteredAccounts table

Elyniss:
According to specs we need to specify `app` with each `wc_notifyWatchSubscriptions`. By specifying the app we limit server to only return us subscriptions matching app field. When app field is null it returns all subscriptions for given account


protocol/notify/src/main/kotlin/com/walletconnect/notify/common/model/RegisteredAccount.kt
```diff
@@ -7,4 +7,5 @@ data class RegisteredAccount(
     val accountId: AccountId,
     val publicIdentityKey: PublicKey,
     val isLimited: Boolean,
+    val appDomain: String?
```
jakubuid, line 10: Who defines it?

Elyniss: In the `NotifyClient.register` developer can specify those parameters

TalhaAli00 (approved):

jakubuid (approved):

PR #1201, opened 2023-10-17
Feature/update sqldelight

TalhaAli00:
This will be merged after 1.18.0


gradle/libs.versions.toml
```diff
@@ -0,0 +1,16 @@
+[versions]
```
jakubuid, line 1: Curious: what doest this file?

jakubuid (approved):
Just to double check - did you check that it doesn't break anything?  E.g. the update path

gradle/libs.versions.toml
TalhaAli00: documentation regarding version catalog
https://developer.android.com/build/migrate-to-catalogs

Elyniss (approved):
I think if we update the branch should be ready to merge

BufferAngel (commented):
I have no idea what  this is?   Or why it is?  What does it do?  Who are you?  Why is this on my phone?

PR #1205, opened 2023-10-18
[W3M] Add requests to test in modal sample and remove session topic from Web3Modal methods

kacperoak:


product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Modal.kt
```diff
@@ -32,10 +32,8 @@ object Modal {
         data class Ping(val topic: String) : Params()
 
         data class Request(
-            val sessionTopic: String,
             val method: String,
             val params: String,
-            val chainId: String,
```
jakubuid, line 38: Why no sessionTopic and chainId in Request anymore?

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Session.kt
```diff
@@ -0,0 +1,23 @@
...
+
+internal object Session {
+
+    private val getSessionTopicUseCase: GetSessionTopicUseCase = wcKoinApp.koin.get()
+    private val getSelectedChainUseCase: GetSelectedChainUseCase = wcKoinApp.koin.get()
+    private val deleteSessionDataUseCase: DeleteSessionDataUseCase = wcKoinApp.koin.get()
+
+    fun getSessionTopic() = getSessionTopicUseCase()
+
+    fun getSelectedChainId() = getSelectedChainUseCase()
```
jakubuid, line 18: Can we add return types to these methods?

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Web3Modal.kt
```diff
@@ -122,7 +120,7 @@ object Web3Modal {
         SignClient.setDappDelegate(signDelegate)
     }
 
-    fun connect(
+    internal fun connect(
```
jakubuid, line 123: What's the reason of making this internal?

sample/modal/src/main/kotlin/com/walletconnect/sample/modal/ui/LabScreen.kt
```diff
@@ -1,24 +1,65 @@
...
     val web3ModalState = rememberWeb3ModalState(navController = navController)
+    val isConnected by web3ModalState.isConnected.collectAsState()
+
+    val coroutineScope = rememberCoroutineScope()
+    val uriHandler = LocalUriHandler.current
+    val context = LocalContext.current
+
+    LaunchedEffect(Unit) {
+        ModalSampleDelegate.wcEventModels.collect {
+            when(it) {
```
jakubuid, line 48: Can we use a name instead of `it`?

sample/modal/src/main/kotlin/com/walletconnect/sample/modal/ui/LabScreen.kt
```diff
@@ -1,24 +1,65 @@
...
+                is Modal.Model.SessionRequestResponse -> {
+                    when(it.result) {
+                        is Modal.Model.JsonRpcResponse.JsonRpcError -> {
+                            val error = it.result as Modal.Model.JsonRpcResponse.JsonRpcError
+                            navController.openAlertDialog("Error Message: ${error.message}\n Error Code: ${error.code}")
+                        }
+                        is Modal.Model.JsonRpcResponse.JsonRpcResult -> navController.openAlertDialog((it.result as Modal.Model.JsonRpcResponse.JsonRpcResult).result)
+                    }
+                }
+                is Modal.Model.Error -> { navController.openAlertDialog(it.throwable.localizedMessage ?:  "Something goes wrong") }
```
jakubuid, line 58:
```suggestion
                is Modal.Model.Error -> { navController.openAlertDialog(it.throwable.localizedMessage ?:  "Something went wrong") }
```

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Modal.kt
kacperoak: because session topic and chain id are data what web3modal keeps in its storage. To send a request, I would have to add getters method for those two to make developer just to pass it back to me. Since the modal controls the session and the chain selection, we don't have to pass it externally to call this method. If there is ever a feature request for such functionality, we can restore it

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Web3Modal.kt
kacperoak: There is no reason to call connect if you using web3modal. We are using it under the hood

jakubuid (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Session.kt
```diff
@@ -0,0 +1,23 @@
...
+import com.walletconnect.web3.modal.domain.usecase.DeleteSessionDataUseCase
+import com.walletconnect.web3.modal.domain.usecase.GetSelectedChainUseCase
+import com.walletconnect.web3.modal.domain.usecase.GetSessionTopicUseCase
+import kotlinx.coroutines.launch
+
+internal object Session {
+
+    private val getSessionTopicUseCase: GetSessionTopicUseCase = wcKoinApp.koin.get()
+    private val getSelectedChainUseCase: GetSelectedChainUseCase = wcKoinApp.koin.get()
+    private val deleteSessionDataUseCase: DeleteSessionDataUseCase = wcKoinApp.koin.get()
```
TalhaAli00, line 14: let's make these lazy so that we don't run into an issue where Koin is not setup by the time this object is instantiated

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Session.kt
TalhaAli00, line 1: Do we need this file? Could you not just inject the usecases into whatever class needs them like Web3Modal.kt?

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Web3Modal.kt
TalhaAli00, line 1: With this file slowly growing with logic, maybe we can copy the `Client`/`Protocol`/`Engine` pattern we have with the other SDKs to break up the public facing functions and the business logic use-cases?

kacperoak: Yea. You aree right. The idea of creating a sepearte Session file was to separate the usecase logic from the client. But maybe I should really go with the `Client/Protocol/Engine` pattern

kacperoak: Todo added

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/client/Session.kt
kacperoak: Session removed and usecased move to Web3Modal.kt

TalhaAli00 (approved):

Elyniss:
Fix for instrumented tests is coming in: https://github.com/WalletConnect/WalletConnectKotlinV2/pull/1216/files#diff-4073215740ba87a6cb3b4a22f07efcdd9842ec3ead2d843bbc056a247d443dc4.

Should be good to merge without it

PR #1207, opened 2023-10-19
[W3M] Modal ui improvements and fixes

kacperoak:


product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/button/ConnectButton.kt
```diff
@@ -47,30 +46,25 @@ internal fun ConnectButton(
...
-                }
-                TextButton(
-                    text = text,
-                    style = ButtonStyle.MAIN,
-                    size = buttonSize,
-                    isEnabled = isEnabled,
-                    onClick = onClick
-                )
+        if (isLoading && isEnabled) {
+            ImageButton(
```
TalhaAli00, line 50: does this need to be an image button since we're not using the onClick parameter?

core/modal/src/main/kotlin/com/walletconnect/modal/ui/components/qr/WalletConnectQRCode.kt
```diff
@@ -24,45 +29,66 @@ import com.walletconnect.modal.ui.ComponentPreview
 import com.walletconnect.modalcore.R
 import androidx.compose.ui.graphics.Color as ComposeColor
 
+enum class QrCodeType {
+    WCM, W3M
```
jakubuid, line 33: Why do we need two QR code types?

jakubuid (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/button/ConnectButton.kt
kacperoak: we don't have to, but then I would have to duplicate the colors and styles logic that we have in the StyledButton under ImageButton component

core/modal/src/main/kotlin/com/walletconnect/modal/ui/components/qr/WalletConnectQRCode.kt
kacperoak: because there are slight differences between the QR style in these 2 products. To be precise, this is the main difference is squre corners radius:

Screenshot 2023-10-20 at 13 41 08
Screenshot 2023-10-20 at 13 40 45

TalhaAli00 (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/components/button/ConnectButton.kt
TalhaAli00: Maybe something to work towards in the future so things that don't need to be a button aren't

kacperoak: I will think about if i can exclude it to separate composable or functions

PR #1208, opened 2023-10-19
[Notify/W3W Sample] Always display PN and handle duplicates

Elyniss:


sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/ui/Web3WalletViewModel.kt
```diff
@@ -17,18 +21,28 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
+import kotlinx.coroutines.flow.filterIsInstance
+import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.merge
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.launch
 
-class Web3WalletViewModel : ViewModel() {
+class Web3WalletViewModel(_app: Application) : AndroidViewModel(_app) {
```
TalhaAli00, line 32: instead of injecting the application and calling `NotificationHandler` in the VM, I think it would be easy to just listen to the event from the ViewModel and call the handler from the view

protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/calls/DecryptMessageUseCase.kt
```diff
@@ -12,13 +12,16 @@ import com.walletconnect.android.internal.common.storage.JsonRpcHistory
...
+import com.walletconnect.notify.data.storage.MessagesRepository
 import kotlinx.coroutines.supervisorScope
+import timber.log.Timber
 import kotlin.reflect.safeCast
 
 internal class DecryptMessageUseCase(
     private val codec: Codec,
     private val serializer: JsonRpcSerializer,
     private val jsonRpcHistory: JsonRpcHistory,
+    private val messagesRepository: MessagesRepository
```
jakubuid, line 24: Can you explain what's the role of MessagesRepository and how it differs from JsonRpcHistory?

Elyniss: MessagesRepository handles storage of Notify Messages, namely `wc_notifyMessage` RPC method. In near future we will add there read statuses etc. It's only inside Notify module. On the other hand, JsonRpcHistory stores all relay messages in Core module

jakubuid: Can't we use JsonRpcHistory to detect duplications?

Elyniss: When message comes from FCM we are not adding entry into JsonRpcHistory

jakubuid: Maybe we should? We're already injecting JsonRpcHistory to DecryptMessageUseCase. `wc_notifyMessage` is also a RPC method. Do we really need a separate storage for this request?

kacperoak (approved):

TalhaAli00 (approved):

PR #1216, opened 2023-10-26
[Notify] Add JWT validation

Elyniss:


jakubuid (approved):

protocol/notify/src/main/kotlin/com/walletconnect/notify/data/jwt/NotifyJwtBase.kt
TalhaAli00, line 1: let's make `ACTION_CLAIM_VALUE` an abstract property, then override it in each child. Then you can extract `throwIfActionIsInvalid` and `throwIfBaseIsInvalid` to JwtBase instead of duplicating the functions in each child Claim class

This will also make sure all children have the exact same behavior

TalhaAli00 (changes requested):

TalhaAli00 (approved):

Elyniss:
IT passes locally

PR #1218, opened 2023-10-26
[W3W Sample] Add url redirection on PN

Elyniss:
Merge after https://github.com/WalletConnect/WalletConnectKotlinV2/pull/1217

### DIFF
--- a/sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/domain/NotificationHandler.kt
+++ b/sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/domain/NotificationHandler.kt
@@ -104,7 +104,7 @@ object NotificationHandler {
                     Triple(notification.messageId, notification.title, notification.body)
                 }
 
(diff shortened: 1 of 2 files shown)


sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/domain/NotificationHandler.kt
```diff
@@ -106,6 +104,9 @@ object NotificationHandler {
                     Triple(notification.messageId, notification.title, notification.body)
                 }
 
+                val url = (notifications.first() as? Notification.Decrypted)?.url
```
TalhaAli00, line 107: Don't we want something like this instead where we want the first instance of a decrypted notification instead of if the first message is a decrypted notification?

```suggestion
val url = (notifications.first { it is Notification.Decrypted }?.url
```

TalhaAli00 (approved):

jakubuid (approved):

kacperoak (approved):

PR #1225, opened 2023-11-02
Feature/meta/add signing config plugin

TalhaAli00:


buildSrc/build.gradle.kts
```diff
@@ -9,8 +9,11 @@ repositories {
     gradlePluginPortal()
 }
 
+kotlin.jvmToolchain(11)
+
```
Elyniss, line 13: This one is always an issue when I try to build my code. Could it be because I use JDK version 18? What's yours?

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -21,7 +21,7 @@ val jvmVersion = JavaVersion.VERSION_11
 const val MIN_SDK: Int = 23
 const val TARGET_SDK: Int = 33
 const val COMPILE_SDK: Int = TARGET_SDK
-val SAMPLE_VERSION_CODE = BOM_VERSION.split(".").run { "${this[1]}${this[2]}" }.toInt()
+val SAMPLE_VERSION_CODE = BOM_VERSION.replace(".", "").toInt() // maybe use Github Actions build#
```
Elyniss, line 24: Not sure if GHA build is incremental

buildSrc/build.gradle.kts
TalhaAli00: I'm using JDK 20. I'm only able to get my code to work with this set

buildSrc/src/main/kotlin/Versions.kt
TalhaAli00: we're actually using the `GITHUB_RUN_ATTEMPT` which will increment with each run of the same build. I'll remove the comment

jakubuid (approved):

sample/wallet/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ plugins {
     kotlin("android")
     id("com.google.gms.google-services")
     id("com.google.firebase.crashlytics")
-    id("com.google.firebase.appdistribution")
+    id("signing-config")
 //    id("io.sentry.android.gradle") version "3.12.0"
```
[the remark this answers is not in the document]
TalhaAli00: I want to leave it and enable it again for internal builds

kacperoak (approved):

PR #1229, opened 2023-11-07
Allow to initialize web3wallet and web3Modal sdks

kacperoak:

PR #1236, opened 2023-11-09
[W3M] Fix delegate changes

kacperoak:


product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/client/WalletConnectModal.kt
```diff
@@ -64,7 +64,7 @@ object WalletConnectModal {
         _sessionParams = init.sessionParams
         runCatching {
             wcKoinApp.modules(walletConnectModalModule())
-            setDelegate(WalletConnectModalDelegate)
+            setInternalDelegate(WalletConnectModalDelegate)
```
kacperoak, line 67: In this [PR](https://github.com/WalletConnect/WalletConnectKotlinV2/pull/1229), I changed the way the developer sets the delegate, but the internal delegate was not set because of it. That's why now I have a separate method that sets the internal delegate on the sign client and the developer sets to only receives evens from Web3ModalDelegegate

TalhaAli00 (approved):

Elyniss (approved):

PR #1249, opened 2023-11-28
fix: Schema koin declaration

Elyniss:
Fixes wrong koin declaration and improves notification displaying within sample app

### DIFF
--- a/protocol/sign/src/main/kotlin/com/walletconnect/sign/di/StorageModule.kt
+++ b/protocol/sign/src/main/kotlin/com/walletconnect/sign/di/StorageModule.kt
@@ -5,6 +5,7 @@ package com.walletconnect.sign.di
 import com.walletconnect.android.di.sdkBaseStorageModule
 import com.walletconnect.android.internal.common.di.AndroidCommonDITags
 import com.walletconnect.android.internal.common.di.deleteDatabase
+import com.walletconnect.foundation.util.Logger
 import com.walletconnect.sign.SignDatabase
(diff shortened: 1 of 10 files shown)


protocol/sign/src/main/kotlin/com/walletconnect/sign/di/StorageModule.kt
```diff
@@ -22,7 +22,7 @@ import org.koin.dsl.module
 @JvmSynthetic
 internal fun storageModule(dbName: String): Module = module {
     fun Scope.createSignDB(): SignDatabase = SignDatabase(
-        get(),
```
kacperoak, line 25: lets add it as named argument here to match the rest

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/Web3WalletApplication.kt
```diff
@@ -197,4 +207,31 @@ class Web3WalletApplication : Application() {
         )
         addFirebaseBeagleModules()
     }
+
+
+    private fun handleNotifyMessages() {
+        val scope = CoroutineScope(Dispatchers.Default)
+
+        val job = NotifyDelegate.notifyEvents
```
kacperoak, line 215: Minor: Can we change variable names  from job and job2 to something more meaningful

kacperoak (approved):

jakubuid (approved):

PR #1253, opened 2023-11-30
[W3M] Regenerate pairing on each connection

kacperoak:


product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/routes/connect/ParingController.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun connect(
+        sessionParams: Modal.Params.SessionParams,
+        onSuccess: (uri: String) -> Unit,
+        onError: (Throwable) -> Unit
+    )
+
+    val uri: String
+}
+
+class PairingControllerImpl: ParingController {
```
TalhaAli00, line 19: let's make this internal

TalhaAli00 (approved):

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/ui/routes/connect/ParingController.kt
```diff
@@ -0,0 +1,55 @@
+package com.walletconnect.web3.modal.ui.routes.connect
+
+import com.walletconnect.android.Core
+import com.walletconnect.android.CoreClient
+import com.walletconnect.web3.modal.client.Modal
+import com.walletconnect.web3.modal.client.Web3Modal
+
+internal interface ParingController {
```
jakubuid, line 8: We already have PairingInterface and PairingControllerInterface in android-core, maybe you can use them?

kacperoak: Only name is similar. I think its better to have separate implementation and interface for this usecase. I can change name if you have any suggestions.

PR #1257, opened 2023-12-07
fix: web3wallet initialization check

jakubuid:


product/web3wallet/src/main/kotlin/com/walletconnect/web3/wallet/client/Web3Wallet.kt
```diff
@@ -299,8 +300,9 @@ object Web3Wallet {
                     while (true) {
                         if (clientInitCounter == 2) {
                             onSuccess()
-                            break
+                            return@withTimeout
                         }
+                        delay(100)
```
Elyniss, line 305: why delay?

Elyniss: ohh nvm

Elyniss (approved):

TalhaAli00 (approved):

jasonwyatt:
This PR doesn't represent a fix to the issue. The crux of the problem is that `clientInitCounter` can _never_ change within the body of the `validateInitializationCount` method, so it's pointless to check it in a loop.

PR #1261, opened 2023-12-08
[Notify] feat: API refactor and deprecation

Elyniss:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/crypto/kmr/BouncyCastleKeyManagementRepository.kt
```diff
@@ -60,6 +61,14 @@ internal class BouncyCastleKeyManagementRepository(private val keyChain: KeyStor
         return PublicKey(publicKey.bytesToHex().lowercase())
     }
 
+    override fun deriveAndStoreEd25519KeyPair(privateKey: PrivateKey): PublicKey {
```
jakubuid, line 64: Where does privateKey come from?

protocol/notify/src/main/kotlin/com/walletconnect/notify/common/model/CacaoPayloadWithIdentityPrivateKey.kt
```diff
@@ -0,0 +1,6 @@
+package com.walletconnect.notify.common.model
+
+import com.walletconnect.android.internal.common.signing.cacao.Cacao
+import com.walletconnect.foundation.common.model.PrivateKey
+
+internal data class CacaoPayloadWithIdentityPrivateKey(val payload: Cacao.Payload, val key: PrivateKey)
```
jakubuid, line 6: Same question about privateKey

core/android/src/main/kotlin/com/walletconnect/android/internal/common/crypto/kmr/BouncyCastleKeyManagementRepository.kt
Elyniss: We generate privateKey in the SDK. Return it to the developer, and later they need to provide it, because we validate cacao payload with public key derived from it

jakubuid (approved):

PR #1264, opened 2023-12-13
[WCM] Migrate to new W3M API

kacperoak:


product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/domain/dataStore/WalletDataStore.kt
```diff
@@ -0,0 +1,167 @@
...
+    private suspend fun fetchInstalledAndRecommendedWallets() = getWalletsUseCase(
+        sdkType = WCM_SDK,
+        page = 1,
+        includes = getPriorityWallets(),
+        excludeIds = WalletConnectModal.excludedWalletsIds
+    )
+
+    suspend fun fetchMoreWallets() {
+        if (searchPhrase.isEmpty()) {
+            if (walletsListingData.wallets.size < walletsListingData.totalCount) {
```
jakubuid, line 84: Can we aggregate these to ifs into one?

jakubuid (approved):

product/walletconnectmodal/src/main/kotlin/com/walletconnect/wcmodal/domain/dataStore/WalletDataStore.kt
Elyniss: I think it should stay separate

Elyniss (approved):

PR #1278, opened 2024-01-19
[W3M] Migrate session to flows and save session on wc scope instead viewmodel

kacperoak:


product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/domain/SessionRepository.kt
```diff
@@ -23,19 +18,19 @@ internal class SessionRepository(
...
 
-    val sessionTopic: StateFlow<String?> = context.sessionStore.data
+    val sessionTopic: Flow<String?> = context.sessionStore.data
         .map { preferences ->
             preferences[SESSION_TOPIC]
-        }.stateIn(scope, started = SharingStarted.Lazily, null)
+        }
 
-    val selectedChain: StateFlow<String?> = context.sessionStore.data
+    val selectedChain: Flow<String?> = context.sessionStore.data
```
TalhaAli00, line 26: Why make these Flows?

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/domain/SessionRepository.kt
```diff
@@ -23,19 +18,19 @@ internal class SessionRepository(
...
         .map { preferences ->
             preferences[SELECTED_CHAIN]
-        }.stateIn(scope, started = SharingStarted.Lazily, null)
+        }
 
-    fun getSelectedChain(): String? = selectedChain.value
+    suspend fun getSelectedChain(): String? = selectedChain.first()
 
-    fun getSessionTopic(): String? = sessionTopic.value
+    suspend fun getSessionTopic(): String? = sessionTopic.first()
```
TalhaAli00, line 33: maybe use `firstOrNull()` instead?

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/domain/SessionRepository.kt
kacperoak: Because stateflows emits first null when i called values immediately after on success and also I have never used these stateflows as state. I always mapped it somewhere anyway to a different state

product/web3modal/src/main/kotlin/com/walletconnect/web3/modal/domain/SessionRepository.kt
kacperoak: These flows are nullable anyway. Does firstOrNull give me something more here?

TalhaAli00: I guess if `sessionTopic` is nullable, maybe not too much. But it might help protect us in the future in case `sessionTopic` ever changes

TalhaAli00 (approved):

jakubuid (approved):

PR #1290, opened 2024-01-26
feat: Curate explorer listing to only show featured projects

Elyniss:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/explorer/data/network/ExplorerService.kt
```diff
@@ -20,7 +20,8 @@ interface ExplorerService {
         @Query("projectId") projectId: String,
         @Query("entries") entries: Int,
         @Query("page") page: Int,
-        @Query("is_verified") isVerified: Boolean,
+        @Query("isVerified") isVerified: Boolean,
+        @Query("isFeatured") isFeatured: Boolean,
```
TalhaAli00, line 24: Were these updated on purpose from using underscores to camel case?

core/android/src/main/kotlin/com/walletconnect/android/internal/common/explorer/domain/usecase/GetProjectsWithPaginationUseCase.kt
```diff
@@ -6,6 +6,6 @@ import com.walletconnect.android.internal.common.explorer.data.model.Project
 class GetProjectsWithPaginationUseCase(
     private val explorerRepository: ExplorerRepository,
 ) {
-    suspend operator fun invoke(page: Int, entries: Int, isVerified: Boolean): Result<List<Project>> =
-        runCatching { explorerRepository.getProjects(page, entries, isVerified).projects }
+    suspend operator fun invoke(page: Int, entries: Int, isVerified: Boolean, isFeatured: Boolean): Result<List<Project>> =
+        runCatching { explorerRepository.getProjects(page, entries, isVerified, isFeatured).projects.sortedBy { it.order ?: Long.MAX_VALUE } }
```
TalhaAli00, line 10: With the fallback to max value, items that don't have an order will be moved to the top. Is that the desired effect? I would imagine they get moved down

TalhaAli00 (approved):

jakubuid (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/explorer/domain/usecase/GetProjectsWithPaginationUseCase.kt
kacperoak: I think this fallback is ok. Max value will put them at the end of the list, which seems correct for me

kacperoak (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/explorer/data/network/ExplorerService.kt
Elyniss: yes, saw it on browser networking calls, and decided to change to match them

PR #1293, opened 2024-01-29
feat: Do not fetch history from remote when end was reached once

Elyniss:


protocol/notify/src/main/kotlin/com/walletconnect/notify/common/model/Subscription.kt
```diff
@@ -23,5 +23,7 @@ internal sealed class Subscription {
         val dappMetaData: AppMetaData? = null,
         val requestedSubscriptionId: Long? = null,
         val relay: RelayProtocolOptions = RelayProtocolOptions(),
+        val idOfLastNotification: String? = null,
```
jakubuid, line 26:
```suggestion
        val lastNotificationId: String? = null,
```

jakubuid (approved):

PR #1295, opened 2024-01-30
fix: improvements for reported issues

jakubuid:
- Add try/catch for registering network callback and emit isAvailable true when an error is caught
- Execute database check for Sign and Core on Dispatcher.IO
- Add custom toString for JsonAdapterEntry


core/android/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
```diff
@@ -68,10 +69,13 @@ fun baseStorageModule(storagePrefix: String = String.Empty) = module {
 
     single<ColumnAdapter<Validation, String>>(named(AndroidCommonDITags.COLUMN_ADAPTER_VALIDATION)) { EnumColumnAdapter() }
 
-    @Suppress("RemoveExplicitTypeArguments")
     single<AndroidCoreDatabase>(named(AndroidBuildVariantDITags.ANDROID_CORE_DATABASE)) {
         try {
-            createCoreDB().also { database -> database.jsonRpcHistoryQueries.selectLastInsertedRowId().executeAsOneOrNull() }
+            createCoreDB().also { database ->
+                com.walletconnect.android.internal.common.scope.launch {
```
kacperoak, line 75: Can we import it? Maybe with alias?

core/android/src/main/kotlin/com/walletconnect/android/internal/common/connection/ConnectivityState.kt
```diff
@@ -48,14 +48,18 @@ internal class ConnectivityState(context: Context) {
...
+                NetworkRequest.Builder()
+                    .addCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+                    .addCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
+                    .addTransportType(NetworkCapabilities.TRANSPORT_CELLULAR)
+                    .addTransportType(NetworkCapabilities.TRANSPORT_WIFI)
+                    .build(),
+                callback
+            )
+        } catch (e: Exception) {
+            _isAvailable.value = true
```
kacperoak, line 62: Should we set true on each exception?

jakubuid: That's the plan because we don't want to block users

core/android/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
jakubuid: Yeah will try with alias

core/android/src/main/kotlin/com/walletconnect/android/internal/common/connection/ConnectivityState.kt
kacperoak: ok, I thought we didn't want to block, in the case of securityException that we talked about recently, but in every other exception we at least want to log it

kacperoak (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/connection/ConnectivityState.kt
jakubuid: Hmm yeah, I forgot about the logs. Will add

protocol/sign/src/main/kotlin/com/walletconnect/sign/di/StorageModule.kt
```diff
@@ -58,7 +59,11 @@ internal fun storageModule(dbName: String): Module = module {
 
     single {
         try {
-            createSignDB().also { signDatabase -> signDatabase.sessionDaoQueries.lastInsertedRow().executeAsOneOrNull() }
+            createSignDB().also { signDatabase ->
+                com.walletconnect.android.internal.common.scope.launch {
+                    signDatabase.sessionDaoQueries.lastInsertedRow().executeAsOneOrNull()
+                }
+            }
```
Elyniss, line 66: should I also add it to Notify and other SDKs?

TalhaAli00: Can you also make sure that if an exception is thrown from the launch that it will get caught by the catch statement

core/android/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
TalhaAli00: Can you also make sure that the catch captures an exception if it's thrown inside of the launch?

TalhaAli00 (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/di/BaseStorageModule.kt
jakubuid: Nice catch

protocol/sign/src/main/kotlin/com/walletconnect/sign/di/StorageModule.kt
jakubuid: Yes, @Elyniss

Elyniss (approved):

kacperoak (approved):

PR #1300, opened 2024-02-01
Sign2.5/Dynamic Switcher

jakubuid:


protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/requests/OnSessionProposalUseCase.kt
```diff
@@ -42,6 +43,10 @@ internal class OnSessionProposalUseCase(
     suspend operator fun invoke(request: WCRequest, payloadParams: SignParams.SessionProposeParams) = supervisorScope {
         val irnParams = IrnParams(Tags.SESSION_PROPOSE_RESPONSE, Ttl(fiveMinutesInSeconds))
         try {
+            if (pairingController.getPairingByTopic(request.topic)?.methods?.contains(JsonRpcMethod.WC_SESSION_AUTHENTICATE) == true) {
+                logger.error("Session proposal received error: pairing supports authenticated sessions")
```
Elyniss, line 47: should this emit some kind of SDKError to _event? Also what about other onErrors in this try catch?

core/android/src/main/kotlin/com/walletconnect/android/internal/common/storage/pairing/PairingStorageRepository.kt
```diff
@@ -79,8 +80,8 @@ class PairingStorageRepository(private val pairingQueries: PairingQueries) : Pai
             relayProtocol = relay_protocol,
             relayData = relay_data,
             uri = uri,
-            registeredMethods = methods,
-            isProposalReceived = is_proposal_received ?: true
+            isProposalReceived = is_proposal_received ?: true,
```
TalhaAli00, line 83: shouldn't this default to false?

core/android/src/main/kotlin/com/walletconnect/android/pairing/handler/PairingController.kt
```diff
@@ -29,6 +29,17 @@ internal class PairingController(private val koinApp: KoinApplication = wcKoinAp
...
     }
 
+    @Throws(IllegalStateException::class)
+    override fun getPairingByTopic(topic: Topic): Pairing? {
+        checkEngineInitialization()
+
+        return try {
+            pairingEngine.getPairingByTopic(topic)
+        } catch (e: Exception) {
+            null
```
TalhaAli00, line 39: let's add a null statement

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignInterface.kt
```diff
@@ -56,7 +56,7 @@ interface SignInterface {
         onError: (Sign.Model.Error) -> Unit
     )
 
-    fun sessionAuthenticate(authenticate: Sign.Params.Authenticate, onSuccess: () -> Unit, onError: (Sign.Model.Error) -> Unit)
+    fun sessionAuthenticate(authenticate: Sign.Params.Authenticate, onSuccess: (String) -> Unit, onError: (Sign.Model.Error) -> Unit)
```
TalhaAli00, line 59: just be aware that changing this will throw an error for all developers who update. It might be best to add a comment about the function and the return type. Maybe we make it a class instead

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -149,15 +153,34 @@ class SignProtocol(private val koinApp: KoinApplication = wcKoinApp) : SignInter
...
+                    emptyMap(),
+                    optionalNamespaces,
+                    properties = null,
+                    pairing = pairing.toPairing(),
+                    onSuccess = {
+                        println("kobe: Proposal Success")
+                        /*Success*/
+                    },
+                    onFailure = { error -> onError(Sign.Model.Error(error)) }
+                )
```
TalhaAli00, line 183: all of this could be in a new use case instead of adding the logic here

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/GetNamespacesFromReCaps.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.walletconnect.android.internal.common.model.Namespace
+import com.walletconnect.android.internal.utils.CoreValidator
+import com.walletconnect.sign.common.validator.SignValidator
+
+class GetNamespacesFromReCaps : GetNamespacesFromReCapsInterface {
+    override fun getNamespacesFromReCaps(chains: List<String>, methods: List<String>): Map<String, Namespace.Proposal> {
+        if (!chains.all { chain -> CoreValidator.isChainIdCAIP2Compliant(chain) }) throw Exception("Chains are not CAIP-2 compliant")
+        if (!chains.any { chain -> SignValidator.getNamespaceKeyFromChainId(chain) == "eip155" }) throw Exception("Only eip155 is supported")
+        val namespace = SignValidator.getNamespaceKeyFromChainId(chains.first())
+        return mapOf(namespace to Namespace.Proposal(events = listOf(), methods = methods, chains = chains))
```
TalhaAli00, line 12: maybe we create a value class for Namespace instead of using String everywhere

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/Pairing.kt
```diff
@@ -13,27 +13,27 @@ data class Pairing(
     val relayProtocol: String,
     val relayData: String?,
     val uri: String,
-    val registeredMethods: String,
     val isProposalReceived: Boolean = false,
+    val methods: String? = null,
```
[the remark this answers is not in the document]
jakubuid: String should be fine. We use this to check if a given peer supports some method. So List IMO is not any better. Wdyt?

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/WalletConnectUri.kt
```diff
@@ -9,8 +9,8 @@ data class WalletConnectUri(
     val symKey: SymmetricKey,
     val relay: RelayProtocolOptions,
     val version: String = "2",
-    val expiry: Expiry?
-//    val registeredMethods: String, // TODO: We'll review later how we want to handle registered methods
+    val expiry: Expiry?,
+    val methods: String?,
```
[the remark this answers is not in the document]
jakubuid: Same, we parse methods from pairing URI where methods are comma separated. I don't see much benefits of having a list

core/android/src/main/kotlin/com/walletconnect/android/pairing/model/mapper/PairingMapper.kt
```diff
@@ -35,7 +35,7 @@ fun Core.Model.Pairing.toPairing(): Pairing =
         relayProtocol,
         relayData,
         uri,
-        registeredMethods
+        methods = registeredMethods
```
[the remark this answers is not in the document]
jakubuid: Because it was already there with this name, so my thinking is to keep it instead of deprecating and changing a name

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
```diff
@@ -299,7 +299,7 @@ object Sign {
         ) : Params()
 
         data class Authenticate(
-            val pairingTopic: String,
+            val pairingTopic: String? = null,
```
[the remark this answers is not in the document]
jakubuid: When dapp wants use a new pairing, then it passes null and the new pairing is created under the hood

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -149,15 +153,34 @@ class SignProtocol(private val koinApp: KoinApplication = wcKoinApp) : SignInter
...
             try {
-                signEngine.authenticate(authenticate.toPayloadParams(), authenticate.methods, authenticate.pairingTopic,
-                    onSuccess = { onSuccess() },
+                val pairing = signEngine.getPairingForSessionAuthenticate(authenticate.pairingTopic)
+                val optionalNamespaces = signEngine.getNamespacesFromReCaps(authenticate.chains, authenticate.methods ?: emptyList()).toMapOfEngineNamespacesOptional()
+
+
+                signEngine.authenticate(authenticate.toPayloadParams(), authenticate.methods, pairing.toPairing(),
+                    onSuccess = { url ->
+                        println("kobe: Auth Success")
```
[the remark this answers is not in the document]
jakubuid: Yes!

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -149,15 +153,34 @@ class SignProtocol(private val koinApp: KoinApplication = wcKoinApp) : SignInter
...
+                    },
                     onFailure = { throwable -> onError(Sign.Model.Error(throwable)) })
+
+                signEngine.proposeSession(
+                    emptyMap(),
+                    optionalNamespaces,
+                    properties = null,
+                    pairing = pairing.toPairing(),
+                    onSuccess = {
+                        println("kobe: Proposal Success")
```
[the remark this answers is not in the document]
jakubuid: Yes!

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/ApproveSessionAuthenticateUseCase.kt
```diff
@@ -54,6 +54,7 @@ internal class ApproveSessionAuthenticateUseCase(
                 return@supervisorScope
             }
             //todo: expiry check
+            //todo: check for single chain
```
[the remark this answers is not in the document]
jakubuid: Yes!

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/GetNamespacesFromReCaps.kt
```diff
@@ -0,0 +1,18 @@
+package com.walletconnect.sign.engine.use_case.calls
+
+import com.walletconnect.android.internal.common.model.Namespace
+import com.walletconnect.android.internal.utils.CoreValidator
+import com.walletconnect.sign.common.validator.SignValidator
+
+class GetNamespacesFromReCaps : GetNamespacesFromReCapsInterface {
+    override fun getNamespacesFromReCaps(chains: List<String>, methods: List<String>): Map<String, Namespace.Proposal> {
+        if (!chains.all { chain -> CoreValidator.isChainIdCAIP2Compliant(chain) }) throw Exception("Chains are not CAIP-2 compliant")
+        if (!chains.any { chain -> SignValidator.getNamespaceKeyFromChainId(chain) == "eip155" }) throw Exception("Only eip155 is supported")
```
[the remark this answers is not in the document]
jakubuid: Yes!

core/android/src/main/kotlin/com/walletconnect/android/internal/common/storage/pairing/PairingStorageRepository.kt
jakubuid: No, because when users upgrade all of the existing pairing already received at least one proposal

core/android/src/main/kotlin/com/walletconnect/android/pairing/handler/PairingController.kt
jakubuid: what do you mean?

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignInterface.kt
jakubuid: > just be aware that changing this will throw an error for all developers who update

Why is that?

jakubuid: > It might be best to add a comment about the function and the return type

This function doesn't have a return type

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
jakubuid: IMO this logic should be in SignProtocol because we can easily reuse proposeSession as a fallback without mixing the crypto logic. 
Things like getPairingForSessionAuthenticate and getNamespacesFromReCaps are already extracted to usecases

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/calls/GetNamespacesFromReCaps.kt
jakubuid: I think it's a good idea but I wouldn't do it in this PR. It'd be better to do it in a separate one.

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/Pairing.kt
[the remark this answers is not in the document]
Elyniss: If it's convinent for you I don't mind it at all. Just saw our custom facing models have this as a list

core/android/src/main/kotlin/com/walletconnect/android/pairing/model/mapper/PairingMapper.kt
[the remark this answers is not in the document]
Elyniss: Makes sense

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/requests/OnSessionProposalUseCase.kt
Elyniss: in this try catch you have multiple times when you are logging errors. Might be worth revisting if some of those could be sent as SDKErrors as well. 

E.g.
```
SignValidator.validateProposalNamespaces(payloadParams.requiredNamespaces) { error ->
    logger.error("Session proposal received error: required namespace validation: ${error.message}")
    jsonRpcInteractor.respondWithError(request, error.toPeerError(), irnParams)
    return@supervisorScope
}
```

could emit error like:

```
SignValidator.validateProposalNamespaces(payloadParams.requiredNamespaces) { error ->
    logger.error("Session proposal received error: required namespace validation: ${error.message}")
    jsonRpcInteractor.respondWithError(request, error.toPeerError(), irnParams)
	_events.emit(SDKError("Received proposal had error while validating namespaces: ${error.message}"))
    return@supervisorScope
}
```

Elyniss (approved):
LGTM. I recommend looking into adding SDKError in OnSessionProposalUseCase on check others as well in near future

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/Pairing.kt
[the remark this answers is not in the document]
jakubuid: Yeah but these are different, and don't want to deprecate this public model

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/use_case/requests/OnSessionProposalUseCase.kt
jakubuid: Yeah, but I'm not sure tbh. We can consider this on the team level and do it on separate PR, since we should change it in all places at once.

Elyniss: Agreed

core/android/src/main/kotlin/com/walletconnect/android/internal/common/storage/pairing/PairingStorageRepository.kt
TalhaAli00: But that wouldn't be the case for new users correct?

core/android/src/main/kotlin/com/walletconnect/android/pairing/handler/PairingController.kt
TalhaAli00: typo

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignInterface.kt
TalhaAli00: No, I'm wrong. Since the String type will be optional for developers to use, it won't cause them issues when upgrading.

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
TalhaAli00: We also have cases where we have use cases inside of use cases since they are meant to be composable. Extracting out this logic will keep up in line with keeping all business logic locked behind the engine class

core/android/src/main/kotlin/com/walletconnect/android/internal/common/storage/pairing/PairingStorageRepository.kt
jakubuid: Can you explain the flow that you're talking about, please?
The new user will create new pairings. Current users will upgrade and all current pairings will have flag set to true

core/android/src/main/kotlin/com/walletconnect/android/pairing/handler/PairingController.kt
jakubuid: Log is added in onSessionProposalUseCase where this method is consumed

TalhaAli00 (approved):

PR #1311, opened 2024-02-06
fix: add topic and chainId to Sigm.Model.SessionEvent

xemniz:


Elyniss:
Hey! May I ask what is the usecase for adding those?

xemniz:
> Hey! May I ask what is the usecase for adding those?

Hi! i need to know selected in wallet chain to send requests to wallet with WC. i hold several sessions, with diferent addresses on MM, maybe one with Rainbow for example and if i want to update current chain from event, i dont know for which topic this event came, because in Sign.Model.SessionEvent i only have method name and params (no session topic)

Elyniss:
Thanks for raising this issue! Based on your proposal I made a PR: https://github.com/WalletConnect/WalletConnectKotlinV2/pull/1315.
There has been a lot of other dependencies than needed to be updated and
in order not to break current implementations we mustn't simply change the data class.

xemniz:
thanks!

PR #1324, opened 2024-02-15
Sign 2.5/Final remarks

jakubuid:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/signing/cacao/Utils.kt
```diff
@@ -1,5 +1,6 @@
 package com.walletconnect.android.internal.common.signing.cacao
```
TalhaAli00, line 1: Should we make the extension functions private?

core/android/src/test/kotlin/com/walletconnect/android/internal/common/cacao/CacaoTest.kt
```diff
@@ -54,7 +53,8 @@ internal class CacaoTest {
     fun signHexAndVerifyWithEIP191Test() {
         println(payload.toCAIP222Message(chainName))
         val message = payload.toCAIP222Message(chainName)
-        val signature: Cacao.Signature = cacaoSigner.signHex(Numeric.toHexString(message.toByteArray()), privateKey, SignatureType.EIP191)
+        val signature: Cacao.Signature = cacaoSigner.sign(Numeric.toHexString(message.toByteArray()), privateKey, SignatureType.EIP191)
+        println("Signature: $signature")
```
TalhaAli00, line 57: Let's use logger instead as to not leak info

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
```diff
@@ -174,6 +176,7 @@ class SignProtocol(private val koinApp: KoinApplication = wcKoinApp) : SignInter
         return try {
             runBlocking { signEngine.formatMessage(formatMessage.payloadParams.toCaip222Request(), formatMessage.iss) }
         } catch (error: Exception) {
+            Timber.e("FormatAuthMessage error: $error")
```
TalhaAli00, line 179: Replace with logger?

protocol/sign/src/main/sqldelight/migrations/9.sqm
```diff
@@ -2,4 +2,10 @@
...
 -- CREATE V10 SCHEMA
 
-ALTER TABLE SessionDao ADD COLUMN is_authenticated INTEGER AS Boolean NOT NULL;
\ No newline at end of file
+-- ALTER TABLE SessionDao ADD COLUMN is_authenticated INTEGER AS Boolean NOT NULL;
+
+CREATE TABLE AuthenticateResponseTopicDao (
+   id INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+   pairingTopic TEXT NOT NULL,
+   responseTopic TEXT NOT NULL
```
TalhaAli00, line 10: Should we add a unique constraint for the both of these?

protocol/sign/src/main/sqldelight/migrations/10.sqm
```diff
@@ -1,9 +0,0 @@
--- migrates 10db to 11db
-
--- CREATE V11 SCHEMA
-
-CREATE TABLE AuthenticateResponseTopicDao (
```
TalhaAli00, line 5: Why delete this file when we have 10.db and 11.db

TalhaAli00 (approved):
replaceWith should just be code since the IDE will let people do a swap with the suggestion. Let's remove the sentences and only put replaceWith if there's a direction replacement for it. The sentences inside replaceWith can go inside the deprecation message

jakubuid:
> replaceWith should just be code since the IDE will let people do a swap with the suggestion. Let's remove the sentences and only put replaceWith if there's a direction replacement for it. The sentences inside replaceWith can go inside the deprecation message

Thank you. I'll keep only methods references in the replaceWith

core/android/src/main/kotlin/com/walletconnect/android/internal/common/signing/cacao/Utils.kt
jakubuid: At least as internal. Changing to internal

jakubuid: Actually, cannot change it to internal because Sign module is using these Utils to parse ReCaps

core/android/src/test/kotlin/com/walletconnect/android/internal/common/cacao/CacaoTest.kt
jakubuid: I think it's testing leftover, IMO we don't need it

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/SignProtocol.kt
jakubuid: I used Timber intentionally because using logger will require me to add code to get it from koin
However, if you feel like it's better I can use logger

protocol/sign/src/main/sqldelight/migrations/9.sqm
jakubuid: No, I'd like to see results of dogfooding first

PR #1331, opened 2024-02-27
fix: approve session proposal, receive session disconnect

jakubuid:


protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -358,6 +368,6 @@ internal class SignEngine(
     }
 
     companion object {
-        private const val WATCHER_INTERVAL = 3000L
+        private const val WATCHER_INTERVAL = 5000L
```
TalhaAli00, line 371: what do you think about bumping this value to something like 10 or even 30 sec? I think to save on resources, we could check less frequently. We also have other checks to validate expiry so I think we would be ok having something longer

core/android/src/main/kotlin/com/walletconnect/android/pairing/engine/domain/PairingEngine.kt
```diff
@@ -317,9 +341,9 @@ internal class PairingEngine(
                 }
             }.launchIn(scope)
 
-    private fun resubscribeToPairingFlow() {
+    private fun resubscribeToPairing() {
         try {
-            val pairingTopics = pairingRepository.getListOfPairings().filter { pairing -> pairing.isNotExpired() }.map { pairing -> pairing.topic.value }
+            val pairingTopics = runBlocking { pairingRepository.getListOfPairings().filter { pairing -> pairing.isNotExpired() }.map { pairing -> pairing.topic.value } }
```
TalhaAli00, line 346: why does this need to be blocking? I think we should think about going the other way and launch a coroutine in the init block and make this function a suspend function to run async. We're already launching coroutines inside of this function

TalhaAli00 (approved):
also curious about why you switched the toast code in the sample

jakubuid:
> also curious about why you switched the toast code in the sample

It was causing session request and proposal dialogs to disappear in some corner cases

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
jakubuid: Yeah I agree. I think that checking every 30s is fine

core/android/src/main/kotlin/com/walletconnect/android/pairing/engine/domain/PairingEngine.kt
jakubuid: Yeah, good point. Made this function a suspend function.

PR #1332, opened 2024-02-28
feat: improve dapp's ui for sign 2.5

jakubuid:


sample/dapp/build.gradle.kts
```diff
@@ -54,6 +54,7 @@ dependencies {
 
     implementation("io.insert-koin:koin-androidx-compose:3.4.3")
     implementation("io.coil-kt:coil-compose:2.3.0")
+    implementation("androidmads.library.qrgenearator:QRGenearator:1.0.4")
```
TalhaAli00, line 57: we already have a dependency for QR code generation which is used in the modal modules
```suggestion
    qrCodeGenerator()
```

core/android/src/main/kotlin/com/walletconnect/android/internal/common/signing/cacao/CacaoVerifier.kt
```diff
@@ -10,6 +10,7 @@ class CacaoVerifier(private val projectId: ProjectId) {
 
         SignatureType.EIP191.header, SignatureType.EIP1271.header -> {
             val plainMessage = cacao.payload.toCAIP222Message()
+            println(plainMessage)
```
TalhaAli00, line 13: remove?

TalhaAli00 (approved):

PR #1344, opened 2024-03-14
feat: CapBox

jakubuid:
CapBox


protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/calls/IsRegisteredUseCase.kt
```diff
@@ -23,12 +23,18 @@ internal class IsRegisteredUseCase(
                     !allApps && registeredAccount.appDomain != domain -> false
                     allApps && !registeredAccount.allApps -> false
                     !allApps && registeredAccount.allApps -> false
-                    else -> identitiesInteractor.getAlreadyRegisteredValidIdentity(AccountId(account), Statement.fromBoolean(allApps).content, domain, listOf(identityServerUrl))
+                    else -> identitiesInteractor.getAlreadyRegisteredValidIdentity(
+                        accountId = AccountId(account),
+                        domain = domain,
+                        resources = listOf(identityServerUrl, createAuthorizationReCaps())
```
jakubuid, line 29: Q: Should the authorization ReCaps be added as well on "is already registered" flow?

chris13524: I think so?

jakubuid: That's what I think too

protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/domain/Utils.kt
```diff
@@ -0,0 +1,6 @@
+package com.walletconnect.notify.engine.domain
+
+fun createAuthorizationReCaps(): String {
+    // {"att":{"https://notify.walletconnect.com":{"manage/all-apps-notifications":[{}]}}}
```
TalhaAli00, line 4: Do we still need this line?

TalhaAli00 (approved):

PR #1350, opened 2024-03-19
[Samples] Update sample to handle generic deeplinks

TalhaAli00:


sample/wallet/src/main/res/navigation/nav_graph.xml
```diff
@@ -6,15 +6,18 @@
...
+        android:fitsSystemWindows="true"
+        android:label="Web3WalletActivity" />
 
     <deepLink
         app:action="android.intent.action.VIEW"
         app:uri="kotlin-web3wallet:/wc" />
 
+    <deepLink
+        app:action="android.intent.action.VIEW"
+        app:uri="wc:/{topic}@2" />
```
jakubuid, line 19: Why this is added?

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/ui/Web3WalletActivity.kt
```diff
@@ -133,17 +153,33 @@ class Web3WalletActivity : AppCompatActivity() {
...
+                intent.setData(uri?.toUri())
+            }
+            intent?.dataString?.startsWith("kotlin-web3wallet:/request") == true -> {
+                val uri = intent.dataString?.replace("kotlin-web3wallet:/request", "kotlin-web3wallet://request")
+                intent.setData(uri?.toUri())
+            }
+            intent?.dataString?.startsWith("wc:") == true -> {
+                val uri = "kotlin-web3wallet://wc?uri=" + URLEncoder.encode(intent.dataString, "UTF-8")
+                intent.setData(uri.toUri())
+            }
```
jakubuid, line 169: Why is it needed?

TalhaAli00: since the sample now only supports deep linking with the `kotlin-web3wallet` scheme, the OS does not show the wallet when trying to deep link using the WalletConnect modal. 
This `when` block is to pass the uri to the correct composable screen

sample/wallet/src/main/res/navigation/nav_graph.xml
jakubuid: I think it's a case only on react-dapp, it uses old wcmodal version. On RN and Flutter they can connect via `kotlin-web3wallet:/wc`

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/ui/Web3WalletActivity.kt
jakubuid: Did you test it with Flutter and Android dapps?

sample/wallet/src/main/res/navigation/nav_graph.xml
TalhaAli00: `kotlin-web3wallet:/wc` is still supported. This just adds the old wcModal version support back into the app

sample/wallet/src/main/kotlin/com/walletconnect/sample/wallet/ui/Web3WalletActivity.kt
TalhaAli00: Android dapp works, just tested Flutter and it works as well

jakubuid (approved):

PR #1362, opened 2024-04-05
fix: activity task management

jakubuid:


sample/wallet/src/main/res/navigation/nav_graph.xml
```diff
@@ -12,11 +12,11 @@
 
     <deepLink
         app:action="android.intent.action.VIEW"
-        app:uri="kotlin-web3wallet:/wc" />
+        app:uri="kotlin-web3wallet://wc" />
 
     <deepLink
         app:action="android.intent.action.VIEW"
-        app:uri="wc:/{topic}@2" />
+        app:uri="wc://{topic}@2" />
```
TalhaAli00, line 19: Did this work with the react Dapp with deep linking?

TalhaAli00 (approved):

sample/wallet/src/main/res/navigation/nav_graph.xml
jakubuid: No, good catch but wondering why // doesn't works since it's a valid deep link

PR #1368, opened 2024-04-15
[meta] removing anything within di files for SonarCloud

TalhaAli00:


build.gradle.kts
```diff
@@ -59,7 +59,7 @@ sonar {
                 "sonar.organization" to "walletconnect",
                 "sonar.host.url" to "https://sonarcloud.io",
                 "sonar.gradle.skipCompile" to true,
-                "sonar.coverage.exclusions" to "sample/**"
+                "sonar.coverage.exclusions" to "sample/**,**/di/**"
```
jakubuid, line 62: Can we add more? Like static files that don't require tests? modals since they require ui tests? what else?

TalhaAli00: let's discuss and create a list of exclusions that we want to add

jakubuid (approved):

PR #1370, opened 2024-04-16
[Core] Catch remove keys exception

TalhaAli00:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/crypto/kmr/KeyManagementRepository.kt
```diff
@@ -8,6 +9,7 @@ import com.walletconnect.foundation.common.model.Topic
 
 interface KeyManagementRepository {
     fun setKey(key: Key, tag: String)
+    @Throws(MissingKeyException::class)
```
jakubuid, line 12: Let's add @Throw to all of the methods in this Interface that throws an exception

jakubuid (approved):

core/android/src/main/kotlin/com/walletconnect/android/internal/common/crypto/kmr/KeyManagementRepository.kt
jakubuid: and let's make sure that other exceptions are caught 🙏

PR #1371, opened 2024-04-16
[core] fix koin bundle

TalhaAli00:


core/android/src/main/kotlin/com/walletconnect/android/internal/common/di/CoreNetworkModule.kt
```diff
@@ -53,7 +54,9 @@ fun coreAndroidNetworkModule(serverUrl: String, connectionType: ConnectionType,
         """wc-2/kotlin-${sdkVersion}/android-${Build.VERSION.RELEASE}"""
     }
 
-    single(named(AndroidCommonDITags.BUNDLE_ID)) { androidContext().packageName }
+    single<PackageName>(named(AndroidCommonDITags.BUNDLE_ID)) {
+        PackageName(androidContext().packageName)
```
jakubuid, line 58: - How does it help? I'm not sure how this resolves a crash where Koin doesn't know how to find a definition for BUNDLE_ID?
- Should we provide a default value as empty string?

jakubuid (approved):

PR #1374, opened 2024-04-18
feat:enhance wss errors handling

jakubuid:


core/android/src/main/kotlin/com/walletconnect/android/pairing/engine/domain/PairingEngine.kt
```diff
@@ -275,12 +276,12 @@ internal class PairingEngine(
     }
 
     private fun resubscribeToPairingTopics() {
-        jsonRpcInteractor.isWSSConnectionAvailable
-            .filter { isAvailable: Boolean -> isAvailable }
+        jsonRpcInteractor.wssConnectionState
+            .filter { state: WSSConnectionState -> state is WSSConnectionState.Connected }
```
TalhaAli00, line 280: `.filterIsInstance<WSSConnectionState.Connected>` is shorter and does the same thing

core/android/src/main/kotlin/com/walletconnect/android/relay/ConnectionState.kt
```diff
@@ -0,0 +1,5 @@
+package com.walletconnect.android.relay
```
TalhaAli00, line 1: add `@file:JvmSynthetic`

core/android/src/main/kotlin/com/walletconnect/android/relay/ConnectionState.kt
```diff
@@ -0,0 +1,5 @@
...
+sealed class WSSConnectionState {
```
TalhaAli00, line 2: add `internal`

core/android/src/main/kotlin/com/walletconnect/android/internal/common/exception/CommonWalletConnectExceptions.kt
```diff
@@ -10,8 +10,9 @@ class CannotFindSequenceForTopic(override val message: String?) : WalletConnectE
 class InvalidProjectIdException(override val message: String?) : WalletConnectException(message)
 class UnableToConnectToWebsocketException(override val message: String?) : WalletConnectException(message)
 class ProjectIdDoesNotExistException(override val message: String?) : WalletConnectException(message)
-class NoInternetConnectionException(override val message: String?) : WalletConnectException(message)
-class NoRelayConnectionException(override val message: String?) : WalletConnectException(message)
+open class NoConnectivityException(override val message: String?) : WalletConnectException(message)
```
TalhaAli00, line 13: should we make this a sealed class instead?

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/ConnectionState.kt
```diff
@@ -2,4 +2,4 @@ package com.walletconnect.android.internal.common.model
 
 import com.walletconnect.android.internal.common.model.type.EngineEvent
 
-data class ConnectionState(val isAvailable: Boolean, val throwable: Throwable? = null) : EngineEvent
+data class ConnectionState(val isAvailable: Boolean, val message: String? = null) : EngineEvent
```
TalhaAli00, line 5: why remove throwable for string?

core/android/src/main/kotlin/com/walletconnect/android/relay/RelayClient.kt
```diff
@@ -14,37 +14,76 @@ import com.walletconnect.foundation.network.data.ConnectionController
...
-            .filterIsInstance<Relay.Model.Event.OnConnectionFailed>()
-            .map { error -> error.throwable.toWalletConnectException }
-            .onEach { walletConnectException -> onError(walletConnectException) }
             .launchIn(scope)
     }
 
+    private fun setIsWSSConnectionOpened(event: Relay.Model.Event) {
+        when (event) {
+            is Relay.Model.Event.OnConnectionOpened<*> -> {
+                if (_wssConnectionState.value is WSSConnectionState.Disconnected) {
```
TalhaAli00, line 66: move the `if` statement into the `is` case for all of the cases

product/web3wallet/src/main/kotlin/com/walletconnect/web3/wallet/client/Wallet.kt
```diff
@@ -52,7 +52,7 @@ object Wallet {
     sealed class Model {
         data class Error(val throwable: Throwable) : Model()
 
-        data class ConnectionState(val isAvailable: Boolean) : Model()
+        data class ConnectionState(val isAvailable: Boolean, val message: String? = null) : Model()
```
TalhaAli00, line 55: why not put the exception?

protocol/auth/src/main/kotlin/com/walletconnect/auth/engine/domain/AuthEngine.kt
```diff
@@ -73,9 +73,8 @@ internal class AuthEngine(
     }
 
     fun setup() {
-        jsonRpcInteractor.isWSSConnectionAvailable
-            .onEach { isAvailable -> _engineEvent.emit(ConnectionState(isAvailable)) }
-            .filter { isAvailable: Boolean -> isAvailable }
+        jsonRpcInteractor.wssConnectionState
+            .filter { state: WSSConnectionState -> state is WSSConnectionState.Connected }
```
TalhaAli00, line 77: can use `filterIsInstance<>` to do the same thing

protocol/chat/src/main/kotlin/com/walletconnect/chat/engine/domain/ChatEngine.kt
```diff
@@ -123,9 +123,8 @@ internal class ChatEngine(
     }
 
     fun setup() {
-        jsonRpcInteractor.isWSSConnectionAvailable
-            .onEach { isAvailable -> _events.emit(ConnectionState(isAvailable)) }
-            .filter { isAvailable: Boolean -> isAvailable }
+        jsonRpcInteractor.wssConnectionState
+            .filter { state: WSSConnectionState -> state is WSSConnectionState.Connected }
```
TalhaAli00, line 127: `filterIsInstance`

protocol/notify/src/main/kotlin/com/walletconnect/notify/engine/NotifyEngine.kt
```diff
@@ -99,9 +100,15 @@ internal class NotifyEngine(
...
-            .filter { isAvailable: Boolean -> isAvailable }
+        jsonRpcInteractor.wssConnectionState
+            .onEach { state ->
+                if (state is WSSConnectionState.Disconnected){
+                    _engineEvent.emit(ConnectionState(false, state.message))
+                } else {
+                    _engineEvent.emit(ConnectionState(true))
+                }
+            }
+            .filter { state: WSSConnectionState -> state is WSSConnectionState.Connected }
```
TalhaAli00, line 111: same filter message

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
```diff
@@ -199,6 +199,7 @@ object Sign {
 
         data class ConnectionState(
             val isAvailable: Boolean,
+            val message: String? = null
```
TalhaAli00, line 202: why not use throwable/exception?

protocol/sign/src/main/kotlin/com/walletconnect/sign/engine/domain/SignEngine.kt
```diff
@@ -187,8 +188,8 @@ internal class SignEngine(
     }
 
     fun setup() {
-        jsonRpcInteractor.isWSSConnectionAvailable
-            .filter { isAvailable: Boolean -> isAvailable }
+        jsonRpcInteractor.wssConnectionState
+            .filter { state: WSSConnectionState -> state is WSSConnectionState.Connected }
```
TalhaAli00, line 192: filter refactor

core/android/src/main/kotlin/com/walletconnect/android/relay/ConnectionState.kt
jakubuid: can't make it internal because it's used in RelayClient that must be public

core/android/src/main/kotlin/com/walletconnect/android/internal/common/exception/CommonWalletConnectExceptions.kt
jakubuid: i think we could but having an open class is also fine in this case
we don't use it in any when statement, so I see only little benefit

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/ConnectionState.kt
jakubuid: because the message is less scary and on the ConnectionClosed state there's no exception thrown, there's a message with reason returned from Scarlet. so i think having a message with an exception message or a reason makes more sense

core/android/src/main/kotlin/com/walletconnect/android/relay/RelayClient.kt
jakubuid: yeah good point

product/web3wallet/src/main/kotlin/com/walletconnect/web3/wallet/client/Wallet.kt
jakubuid: answered above

protocol/sign/src/main/kotlin/com/walletconnect/sign/client/Sign.kt
jakubuid: answered above

core/android/src/main/kotlin/com/walletconnect/android/internal/common/exception/CommonWalletConnectExceptions.kt
TalhaAli00: an open class will also let others expand this class. I guess we can leave it for now

core/android/src/main/kotlin/com/walletconnect/android/internal/common/model/ConnectionState.kt
TalhaAli00: should we maybe do both? Like have a message and a throwable field? When there is a real problem, having an exception would be really helpful to debug the issue. Maybe we have a inner class/sealed class to handle both cases

TalhaAli00 (approved):

PR #1375, opened 2024-04-18
[Core] updated web3j dependency

TalhaAli00:


buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -54,7 +54,7 @@ const val mockkVersion = "1.13.5"
 const val jsonVersion = "20220924"
 const val timberVersion = "5.0.1"
 const val androidSecurityVersion = "1.1.0-alpha06"
-const val web3jVersion = "4.9.8-wc"
+const val web3jVersion = "4.9.8-hotfix"
```
jakubuid, line 57: is it the final version?

TalhaAli00: yes

TalhaAli00: All future versions use JDK 17. We won't have support for 17 until we bump our compile version but that comes with it's own problems

jakubuid (approved):

PR #1402, opened 2024-05-22
feat: execution tracing

jakubuid:
Implementation of execution tracing for Events SDK. Error events are stored in the local storage.


jakubuid:
FYI: test will be added in one last PR

PR #1464, opened 2024-07-31
feat: Verify v2

jakubuid:
Sample Debug Kotlin Wallet with Verify V2: https://appdistribution.firebase.google.com/testerapps/1:1056012215045:android:e94d26b8752f3d5c11cd6c/releases/5mvfdg5qn8328?utm_source=firebase-console

Sample JS Dapp with Verify V2: https://react-dapp-v2-git-chore-verify-v2-samples-walletconnect1.vercel.app/

@chris13524 please test on your side as well before it's merged 🙏


core/android/src/main/kotlin/com/walletconnect/android/verify/client/VerifyClient.kt
```diff
@@ -1,41 +1,38 @@
...
+    private val scope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
+) : VerifyInterface {
+    private val verifyRepository by lazy { koinApp.koin.get<VerifyRepository>() }
 
     override fun initialize() {
         koinApp.modules(verifyModule())
+
+        scope.launch {
+            verifyRepository.getVerifyPublicKey().onFailure { throwable -> println("Error fetching a key: ${throwable.message}") }
+        }
```
chris13524, line 24: Does this run async and not block SDK startup?

core/android/src/main/kotlin/com/walletconnect/android/verify/domain/VerifyRepository.kt
```diff
@@ -0,0 +1,114 @@
...
+        }
+        publicKey
+    }
+
+    fun resolveV2(attestationJWT: String, metadataUrl: String, onSuccess: (VerifyResult) -> Unit, onError: (Throwable) -> Unit) {
+        scope.launch {
+            supervisorScope {
+                getVerifyPublicKey()
+                    .fold(
+                        onSuccess = { key ->
```
chris13524, line 37: Does this await the original `getVerifyPublicKey()` that was called in `initialize()`? I.e. maybe we need a lock?

core/android/src/main/kotlin/com/walletconnect/android/verify/client/VerifyClient.kt
jakubuid: It's async, doesn't block the SDK startup

> Furthermore, it should only run when the client has an intention to pair. I.e. we don't want every SDK instance making this request if they aren't using it.
Agree, will change that

core/android/src/main/kotlin/com/walletconnect/android/verify/domain/VerifyRepository.kt
jakubuid: Actually, good point, it would be better to await in this case

PR #1487, opened 2024-08-23
feat: JWT id check

jakubuid:


core/android/src/test/kotlin/com/walletconnect/android/internal/VerifyRepositoryTest.kt
```diff
@@ -93,7 +95,38 @@ class VerifyRepositoryTest {
...
+
+    @Test
+    fun `resolveV2 calls onSuccess when JWT is valid and ids do not match`() = testScope.runTest {
+        val attestationId = "test"
+        val attestationJWT = "attestationJWT"
+        val metadataUrl = "https://metadata.url"
+        val publicKey = "0409b2f80ce60e6f59ed77ef0e984c4efa84b40d608c0b4d039edaf2989a01f2d92931708c7b50c464c347dd55b0eca971d05fbdba3ab00323e69e166fef61440d"
+        val verifyResult = VerifyResult(Validation.INVALID, false, "https://react-dapp-v2-git-chore-verify-v2-samples-walletconnect1.vercel.app")
+        val claimsJson =
+            """{"exp":1722579908,"id":"5106a25552e89acfb5bed83ee21bf4e80dbcd51b0b203f6925a369aacb1c860b","origin":"https://react-dapp-v2-git-chore-verify-v2-samples-walletconnect1.vercel.app","isScam":null,"isVerified":true}"""
```
chris13524, line 113: How does this test work? Are we somehow ignoring the expiration here?

chris13524 (approved):

core/android/src/test/kotlin/com/walletconnect/android/internal/VerifyRepositoryTest.kt
jakubuid: We check against expiration. If JWT is expired the validation is UNKNOWN

In the above test case, we test if the verify v1 is called and resolves when there's ids mismatch. If there's a mismatch the code calls v1 so the jwt expiration check doesn't happen

chris13524: Interesting that expiration check happens after ID check! Cool

